PR #1, opened 2018-03-15
removed whitespace from path

mmacata:
This PR removes white spaces from paths in src/actinia_core/resources/common/config.py, to avoid escaping it in config / init files


neteler:
@ Soeren:
IMHO it makes sense to use DEFAULT_CONFIG_PATH = "/etc/default/actinia_core" (rather than "/etc/default/actinia")...

Issue #7, opened 2018-11-01
Add missing headers to various files


neteler:
- https://github.com/mundialis/actinia_core/blob/master/src/actinia_core/__init__.py
- https://github.com/mundialis/actinia_core/blob/master/src/actinia_core/endpoints.py
...

neteler:
I have added the license headers in  0dacc74, will open a PR

neteler:
All headers updated in 0dacc74

PR #9, opened 2018-11-02
WIP Add file headers for GPLv3

marcjansen:
This is a first draft of fixing #7 

@neteler is this what you have envisioned? If so, this can be added to the other files.


neteler:
Thanks for the suggestion: I have updated all headers in a similar way in 0dacc74

marcjansen:
Cool. Thanks.

PR #17, opened 2019-03-01
remove venv usage for GRASS GIS 7.7

mmacata:
As with grass7.7 comes python3 support, we don't need to use virtual python environments anymore to separate between actinia (python3) and GRASS GIS (python2).
This PR removes the usage of the virtualenv for GRASS GIS


mmacata:
@neteler @AnikaBettge @mundialis-dev are we ready for this? It means that latest actinia can only be installed with GRASS GIS 7.7 using python3.
If deployments are done with GRASS GIS using python2, we need to checkout the[ latest release v0.11](https://github.com/mundialis/actinia_core/releases/tag/v0.1.1)
Please review and tell your opinion

neteler:
To me it looks good, thanks

PR #18, opened 2019-03-29
interface description output

mmacata:
This PR suggests to let the GRASS module interface description be printed via process chain parameter


mmacata:
This PR is WIP. To be discussed, if we really need this

mmacata:
Not WIP anymore, we need it. It is tested and awaiting review once again ;) @AnikaBettge

src/actinia_core/resources/common/process_chain.py
```diff
@@ -989,6 +991,9 @@ def _create_module_process(self, module_descr):
         if "verbose" in module_descr and module_descr["verbose"] is True:
             params.append("--v")
 
+        if "interface-description" in module_descr and module_descr["interface-description"] == "True":
```
marcjansen: Just a question, why is it `"True"` here (a string), but for the `verbose` parameter above it is `True` (a boolean)?

src/actinia_core/resources/common/process_chain.py
```diff
@@ -1284,6 +1290,9 @@ def _create_module_process_legacy(self, id, module_descr):
         if "verbose" in module_descr and module_descr["verbose"] is True:
             parameters.append("--v")
 
+        if "interface-description" in module_descr and module_descr["interface-description"] == "True":
```
marcjansen: See other question.

src/actinia_core/resources/common/process_chain.py
mmacata: Very good hint, there was some process-chain confusion about that value. We found it thanks to your hint and fixed it. I changed above to a boolean

mmacata:
Ok thank you, I'll merge now

PR #21, opened 2019-04-16
module list in existing location

AnikaBettge:
- until now the list is only created for a request with a user


neteler:
Not sure if useful at all: on command line it would look like this

```
# inside a location: reverse search for a non-existing keyword prints all descriptions incl addons:
g.search.modules -n keyword=dummy -j | grep name  | wc -l
531

# from outside a location
grass77 -c epsg:4326 --tmp-location --exec g.search.modules -n keyword=dummy -j | grep name  | wc -l
Starting GRASS GIS...
Creating new GRASS GIS location <tmploc>...
Cleaning up temporary files...
Executing <g.search.modules -n keyword=dummy -j> ...
Execution of <g.search.modules -n keyword=dummy -j> finished.
Cleaning up temporary files...
531
```

mmacata:
As we moved this funcionality to actinia-gdi, I'll close this MR. See https://github.com/mundialis/actinia-gdi/pull/2

PR #23, opened 2019-04-18
Check actinia modules

AnikaBettge:
#6954


mmacata:
Moved to actinia-gdi (https://github.com/mundialis/actinia-gdi/pull/3) and split to https://github.com/mundialis/actinia_core/pull/24. Therefore I close now

PR #25, opened 2019-04-25
overwrite config path if env is set

mmacata:
title says it all ...
additionally to #20 necessary if used in scripts


AnikaBettge:
looks good. you can merge

mmacata:
thx i#ll merge now

PR #27, opened 2019-05-08
switch docker build based on master instead of release

mmacata:
This concerns only the docker build. Before the source was fetched by release tag. Now it is switched to master branch to always build latest code.


neteler:
Look ok to me

mmacata:
Ok, thx for reviewing, I'll merge now

PR #30, opened 2019-05-09
add production docker config and remove GISBASE (fixed in GRASS)

mmacata:
This PR adds a docker-compose-prod.yml for production environments. It also removes the setting of the GISBASE environment variable before starting GRASS GIS (fixed in GRASS GIS itself)


mmacata:
@neteler told me in person that it is fine to merge, so I'll merge now...

PR #32, opened 2019-05-09
Correct an error message text

marcjansen:
The title says it all, the wrong text is probably due to some copy and pasting. Please review.


neteler:
Thanks!

PR #33, opened 2019-05-10
enable strace and gdb within docker container (docker dev)

neteler:
..would be cool to have that enabled for debugging.


neteler:
While for docker it works on command line, it doesn't for docker-compose. Here you need to change the yml file

PR #37, opened 2019-09-01
fix LL location name to latlong_wgs84 (on server)

neteler:


neteler:
Probably an update is also needed in this line?

https://github.com/mundialis/actinia_core/blob/2151de5e1b41a12897f37465678612a4c595f71e/src/actinia_core/testsuite.py#L179

PR #41, opened 2019-09-25
optimize gunicorn startup options

mmacata:
This PR optimizes the startup options with which the python flask server is run (gunicorn).
As read in https://wiki.ubuntuusers.de/Gunicorn/ and http://docs.gunicorn.org/en/stable/design.html, the preferred worker for this setup is gthread, which is an asynchronous worker. For development and easier debugging, only one worker in parallel is started.


mmacata:
As discussed with @neteler and @AnikaBettge, I'll merge now

PR #46, opened 2019-10-08
WIP: start/stop webhook server for tests

neteler:
- start webhook server prior to running tests
- stop after tests


neteler:
Mhh, it is started but not used. Weird...

neteler:
Replaced by #105

PR #49, opened 2019-10-15
make swagger docs of AsyncPersistentResource reusable

mmacata:
As actinia-gdi extends the endpoint for pesistent processing to translate process chain templates to actinia-modules, it would be nice to reuse the swagger docs for this. Therefore persistent_processing.py was adjusted to look like ephemeral_processing_with_export.py.

### DIFF
--- a/src/actinia_core/resources/persistent_processing.py
+++ b/src/actinia_core/resources/persistent_processing.py
@@ -29,6 +29,7 @@ import pickle
 import os
 import shutil
 import subprocess
+from copy import deepcopy
 from flask import jsonify, make_response
 from flask_restful_swagger_2 import swagger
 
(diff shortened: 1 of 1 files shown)


src/actinia_core/resources/persistent_processing.py
```diff
@@ -79,51 +80,54 @@
...
+        }
+    ],
+    'responses': {
+        '200': {
+            'description': 'The result of the process chain execution',
+            'schema':ProcessingResponseModel
+        },
+        '400': {
+            'description':'The error message and a detailed log why process chain execution '
+                          'did not succeeded',
```
neteler:
```suggestion
                          'did not succeed',
```

mmacata: fixed typo

PR #50, opened 2019-10-18
use copy instead of git pull in Dockerfile

mmacata:
This way, it will be possible to test branches and tags or even local changes by simply copying the current source code into the docker instead of specifying branch or tag to checkout from github.


mmacata:
As approved, I'll merge now..

PR #56, opened 2019-10-30
make base dockerimage explicit (grass78)

mmacata:
As we use grass-py3-pdal as baseimage which changed default version in latest image from grass78 (stable) to grass79 (develop version), we need to explicitly set the docker tag in the image.


mmacata:
As approved, I'll merge now

PR #57, opened 2019-10-30
use git tag for actinia version

mmacata:
Before, the version of actinia was hardcoded in src/actinia_core/resources/common/app.py in API_VERSION="v1". It was used to created the api path (/api/v1/) and to display the version for /api/v1/version. Now for the latter, the version will be retrieved from the git version (see `git describe`, "Default versioning scheme" on https://pypi.org/project/setuptools-scm/ ). This might lead to weird looking versions during development but should create nice semver versions for production.
Example (during development):
```
{
  "plugin_version": {
    "actinia_gdi": "0.1.1.post0.dev8+g5f2c551.dirty"
  },
  "plugins": "actinia_gdi",
  "version": "0.2.2.post0.dev13+g9382e50.dirty"
}
```
After merge I would create a release v0.99.0 to closely align API version (v1 in path), actinia-version, github release tag version and Docker tag version. (not 1.0.0, waiting for https://trac.osgeo.org/grass/ticket/3928)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -5,12 +5,6 @@ services:
     build:
       context: ./..
       dockerfile: docker/actinia-core/Dockerfile
-      args:
-        - SOURCE_GIT_URL=https://github.com
-        - SOURCE_GIT_REMOTE=mundialis
-        - SOURCE_GIT_REPO=actinia_core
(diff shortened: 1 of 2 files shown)


mmacata:
As discussed with @neteler and @AnikaBettge , I'll merge now

PR #58, opened 2019-10-30
fix actinia curl examples

neteler:


neteler:
Thanks for the review, merging and closing.

PR #59, opened 2019-10-31
fix version to not be dirty after release

mmacata:
This PR is a follow up to #57. Instead of the version shown there, after a clean release it is shown like this:

```
{
  "plugins": "actinia_statistic_plugin", 
  "version": "0.99.0+ga8ff003.dirty"
}
```

This "not clean" part are files which are not copied into the docker but commited (some docker files not relevant for the main build, ...)
This PR suggests to conduct a `git pull` to make the version clean again.


mmacata:
Ok, then I'll merge now

PR #67, opened 2019-11-20
actinia docker image based on GRASS GIS 7.9 (master)

neteler:
Separate Dcokerfile to facilitate deployments with GRASS GIS 7.9 (master) included which provides "image collections" support.

Note: temporary solution until multistage images have been designed.


mmacata:
As we discussed that we want to restructure to multistage-build very soon, I approve (even though this PR generates duplicate code)

PR #68, opened 2019-11-20
exporter: support attributes at raster export (GDAL RAT)

neteler:


mmacata:
Looks good, pls merge

PR #78, opened 2020-01-28
Adjust to google_cloud_bigquery update

mmacata:
As we updated our requirements in requirements.txt, google_cloud_bigquery API changed and our code needs to be adjusted accordingly.
See https://github.com/googleapis/google-cloud-python/blob/master/bigquery/CHANGELOG.md#interface-changes--breaking-changes

PR #79, opened 2020-01-29
Remove timeout from gc-bigquery after API change

mmacata:
See https://github.com/mundialis/actinia_core/pull/78, this is also necessary.


neteler:
Looks good, pls merge

PR #81, opened 2020-02-07
Sync grass_addons_list.csv in Dockerfiles

mmacata:
Until now, all Dockerfiles had their own grass_addons_list.csv file with GRASS GIS extensions. This leads to confusion. Now they all share the same one.


mmacata:
As only minor change, I'll merge

PR #83, opened 2020-03-02
Fix cancel/termination operation for redis version >= 3.0.0

AnikaBettge:
There are some changes in redis (https://github.com/andymccurdy/redis-py/blob/master/CHANGES) so that the cancel operation does not work for redis>=2.10.6 (and rq>=0.10.0).
This is caused by the change in 3.0.0: Only bytes, strings and numbers (ints, longs and floats) are acceptablefor keys and values. Previously redis-py attempted to cast other types
to str() and store the result. This caused must confusion and frustration
when passing boolean values (cast to 'True' and 'False') or None values
(cast to 'None'). It is now the user's responsibility to cast all
key names and values to bytes, strings or numbers before passing the
value to redis-py.


mmacata:
Cool, thank you, please merge!

PR #84, opened 2020-03-04
Switch alpine docker to alpine 3.11 and compiled pdal

mmacata:
As alpine:edge leads to uncontrolled package updates which might break things, the use of a stable alpine version as base image is much more stable. Therefore this PR

- uses alpine:3.11 as base image for build and runtime images (actinia-core following as soon as other images from which it depends have switched as well)
- uses compiled PDAL and LASZIP because they do not exist as alpine:3.11 packages
- gets rid of unstable edge-testing package repository
- uses openjdk8-232 for snappy


mmacata:
As all images from which actinia depends are build, I'll merge now.

PR #87, opened 2020-04-09
Messages: fix Sentinel 2A to Sentinel-2

neteler:
... be less confusing since Sentinel-2B also exists.


neteler:
Any objections to merge this?

PR #90, opened 2020-06-03
fix gdal for ubuntu docker

mmacata:
As snappy S2 and S3 Toolbox bring its own GDAL version (2.2.0) which conflicts with system version (and breaks in our code), they are removed as we only use S1TBX anyway.


mmacata:
Thx for reviewing, I'll merge now

PR #91, opened 2020-06-16
update version in actinia-gdi alpine Dockerfile

mmacata:
title says it all..


mmacata:
As only minor change in Dockerfile, I'll merge directly

PR #92, opened 2020-06-17
docker: alpine version 3.12

AnikaBettge:
There is a new alpine version 3.12 (https://wiki.alpinelinux.org/wiki/Alpine_Linux:Releases).
Depends on mundialis/esa-snap:s1tbx-7.0v2 (https://hub.docker.com/r/mundialis/esa-snap/tags) and on grass-py3-pdal:stable-alpine (https://hub.docker.com/repository/docker/mundialis/grass-py3-pdal/builds)

### DIFF
--- a/docker/actinia-core-alpine/Dockerfile
+++ b/docker/actinia-core-alpine/Dockerfile
@@ -74,6 +74,7 @@ COPY --from=build /build/*.whl /build/
 # RUN apk add curl && curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3 get-pip.py pip==20.0.2 && rm get-pip.py
 RUN pip3 install /build/*
 # See in build stage why we need this here
+# TODO: replace edge testing packages as soon as possible.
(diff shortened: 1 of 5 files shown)


docker/actinia-core-alpine/Dockerfile
```diff
@@ -74,7 +74,7 @@ COPY --from=build /build/*.whl /build/
 # RUN apk add curl && curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && python3 get-pip.py pip==20.0.2 && rm get-pip.py
 RUN pip3 install /build/*
 # See in build stage why we need this here
-RUN pip3 install scikit-learn
+RUN apk add --repository http://dl-cdn.alpinelinux.org/alpine/edge/testing py3-scikit-learn
```
mmacata, line 77: In general, adding a package from edge to a stable alpine version (3.12) might lead to errors due to unmet dependencies, but in this case I guess it is fine.

AnikaBettge: Thanks. I added a TODO for this

PR #96, opened 2020-07-22
fix scikit-learn in alpine docker

AnikaBettge:
add py3-joblib from edge community after installing pip requirements


neteler:
I'll merge to hopefully silent the dockerhub failure.

PR #97, opened 2020-07-27
Update requirements.txt and add few tests

AnikaBettge:
Only the `psutil` had a fixed version which should be updated. The version of `docutils` is fixed to 0.15 cause `botocore` requires a `docutils<0.16,>=0.10`.

In the Dockerfile some tests are added.


mmacata:
Looks good. And we get some tests in CI, yeah!

PR #107, opened 2020-09-17
return codes: explanations added

neteler:


marcjansen (approved):
Thanks. Looks good to me.

PR #108, opened 2020-09-17
raster_exporter: COG support added

griembauer:
Raster exporter does not yet support COG, for which GDAL >= 3.1 is required. Current workaround because of GDAL bug in overview creation.

Status: should work for now, but more elegant solution to come after GDAL bugfix


src/actinia_core/resources/ephemeral_processing_with_export.py
```diff
@@ -241,14 +240,29 @@ def _export_raster(self, raster_name,
...
+
+        if format == "GTiff":
+            # generate overviews with compression:
+            os.environ['COMPRESS_OVERVIEW'] = "LZW"
+            args.extend(["createopt=COMPRESS=LZW,TILED=YES", "overviews=5"])
+
+        # DELETE ONCE GDAL 3.2 HAS BEEN RELEASED
+        elif format == "COG":
+            args.append("createopt=OVERVIEWS=NONE")
+        ###
```
AnikaBettge, line 265: Why 3 `#`?

src/actinia_core/resources/ephemeral_processing_with_export.py
```diff
@@ -261,6 +275,14 @@ def _export_raster(self, raster_name,
         self._update_num_of_steps(1)
         self._run_module(p)
 
+        # DELETE ONCE GDAL 3.2 HAS BEEN RELEASED
```
AnikaBettge, line 278: Maybe add a TODO. Should the overview than be set in line 264? Or do we have to test it?

src/actinia_core/resources/ephemeral_processing_with_export.py
```diff
@@ -209,11 +209,12 @@ def _export_raster(self, raster_name,
...
         The region of the raster layer can be used for export. In this case a temporary region
         will be used for export, so that the original region of the mapset is not modified.
+        COG-Driver: https://gdal.org/drivers/raster/cog.html
 
         Args:
             raster_name (str): The name of the raster layer
-            format (str): GTiff
-            additional_options (list): Unused
+            format (str): COG
+            additional_options (list): Additional GDAL create options
```
AnikaBettge, line 217: It is still unused

src/actinia_core/resources/ephemeral_processing_with_export.py
```diff
@@ -199,7 +199,7 @@ def __init__(self, rdc):
         self.storage_interface = rdc.create_storage_interface()
 
     def _export_raster(self, raster_name,
-                       format="GTiff",
+                       format="COG",
```
AnikaBettge, line 202: If i saw it right, there are 2 more endpoints that export raster:
1. /locations/{location_name}/mapsets/{mapset_name}/raster_layers/{raster_name}/geotiff_async
2. /locations/{location_name}/mapsets/{mapset_name}/raster_layers/{raster_name}/geotiff_async_orig
They are also using `_export_raster` in the `AsyncEphemeralRasterLayerExporterResource` calss. So should they really use COG automatically? When they should use it, the docu should be changed:
- https://github.com/mundialis/actinia_core/blob/master/src/actinia_core/resources/raster_export.py#L56
- https://github.com/mundialis/actinia_core/blob/master/src/actinia_core/resources/raster_export.py#L131

src/actinia_core/resources/ephemeral_processing_with_export.py
griembauer: COG sets overviews automatically if not specified differently. Both blocks can be deleted once the GDAL bug is fixed.

src/actinia_core/resources/ephemeral_processing_with_export.py
griembauer: You are right, _export_raster is only used once and there is no additional options argument.

src/actinia_core/resources/ephemeral_processing_with_export.py
griembauer: Both call the EphemeralRasterLayerExporter class, where 'GeoTIFF' was hardcoded as format. The gdal-driver check is now included here such that COG is chosen if the COG driver is available. Docu is adapted accordingly.

metzm:
The bug in COG creation is caused by color table export. The COG driver sets a TIFF color table even if this is not supported by the TIFF specs. With TIFF, color tables can only be set for datatypes Byte and UInt16. Therefore we need to explicitly set the "-c" flag with r.out.gdal when exporting to COG.

src/actinia_core/resources/raster_export.py
```diff
@@ -128,7 +128,7 @@ def __init__(self):
 
     @swagger.doc({
         'tags': ['Raster Management'],
-        'description': 'Export an existing raster map layer as GeoTiff using the raster map layer specific region. '
+        'description': 'Export an existing raster map layer as GeoTiff or COG (if COG driver available) using the raster map layer specific region. '
```
AnikaBettge, line 131: I think the upper text is better. This sounds more like having a choice

AnikaBettge (approved):

src/actinia_core/resources/ephemeral_processing_with_export.py
```diff
@@ -241,14 +239,29 @@ def _export_raster(self, raster_name,
...
-                "createopt=COMPRESS=LZW,TILED=YES", "overviews=5", "output=%s"%output_path]
+        args = ["-fmt", "input=%s" % raster_name, "format=%s" % format, "output=%s" % output_path]
+
+        if format == "GTiff":
+            # generate overviews with compression:
+            os.environ['COMPRESS_OVERVIEW'] = "LZW"
+            args.extend(["createopt=COMPRESS=LZW,TILED=YES", "overviews=5"])
+
+        # current workaround due to color table export
+        # TODO: DELETE AND TEST ONCE GDAL 3.2 HAS BEEN RELEASED
```
neteler, line 262:
```suggestion
        # TODO: DELETE AND TEST ONCE GDAL 3.1.4 HAS BEEN RELEASED
```
See: https://github.com/OSGeo/gdal/issues/2946

src/actinia_core/resources/ephemeral_processing_with_export.py
```diff
@@ -241,14 +239,29 @@ def _export_raster(self, raster_name,
...
-        args = ["-fmt", "input=%s"%raster_name, "format=%s"%format,
-                "createopt=COMPRESS=LZW,TILED=YES", "overviews=5", "output=%s"%output_path]
+        args = ["-fmt", "input=%s" % raster_name, "format=%s" % format, "output=%s" % output_path]
+
+        if format == "GTiff":
+            # generate overviews with compression:
+            os.environ['COMPRESS_OVERVIEW'] = "LZW"
+            args.extend(["createopt=COMPRESS=LZW,TILED=YES", "overviews=5"])
+
+        # current workaround due to color table export
```
metzm, line 261:
```suggestion
        # current workaround due to color table export
        # COG bug in GDAL, see https://github.com/OSGeo/gdal/issues/2946
```

src/actinia_core/resources/ephemeral_processing_with_export.py
metzm, line 262:
```suggestion
        # TODO: DELETE AND TEST ONCE GDAL 3.1.4 HAS BEEN RELEASED
```

src/actinia_core/resources/raster_export.py
```diff
@@ -221,8 +221,13 @@ def _execute(self):
         self.required_mapsets.append(self.mapset_name)
         self._create_temporary_grass_environment(source_mapset_name="PERMANENT")
 
+        format = "COG"
```
metzm, line 224:
```suggestion
        # COG bug in GDAL, see https://github.com/OSGeo/gdal/issues/2946 will be fixed in GDAL 3.1.4
        # use r.out.gdal -c to avoid the bug
        format = "COG"
```

neteler:
I have the impression that COG requires BIGTIFF activation (the implemented auto-detection failed for me several times). And also compression to be activated, as there is none set as the default: https://gdal.org/drivers/raster/cog.html#general-creation-options ).

Any COG > 4GB was corrupted for me (GDAL 3.1.2 on Alpine).

Hence, e.g.:

`r.out.gdal ... format=COG createopt="COMPRESS=LZW,BIGTIFF=YES"`

But with that I get:

```
Warning 6: driver MEM does not support creation option COMPRESS
Warning 6: driver MEM does not support creation option BIGTIFF
```

Does @metzm have an idea?

src/actinia_core/resources/raster_export.py
```diff
@@ -221,8 +221,15 @@ def _execute(self):
         self.required_mapsets.append(self.mapset_name)
         self._create_temporary_grass_environment(source_mapset_name="PERMANENT")
 
+        # COG bug in GDAL, see https://github.com/OSGeo/gdal/issues/2946 will be fixed in GDAL 3.1.4
+        # use r.out.gdal -c to avoid the bug
+        format = "COG"
```
neteler, line 226: See https://github.com/mundialis/actinia_core/pull/108#issuecomment-696174769

griembauer: a r.out.gdal test using `format=COG createopt="BIGTIFF=YES,COMPRESS=LZW"` with a resulting .tiff size of ~12 GB was successful und could also be imported in QGIS without problems.

neteler (approved):

mmacata (approved):

PR #113, opened 2020-09-23
PEP8: fix: 'if cond is not None:', 'if cond is False:' and '3 placeholder(s) but…'

AnikaBettge:
This PR fixes:

- 4     E711 comparison to None should be 'if cond is not None:'
- 1     E712 comparison to False should be 'if cond is False:' or 'if not cond:'
- 1     F507 '...' % ... has 3 placeholder(s) but 5 substitution(s)

### DIFF
--- a/src/actinia_core/resources/common/user.py
+++ b/src/actinia_core/resources/common/user.py
@@ -361,11 +361,6 @@ class ActiniaUser(object):
 
         return(string)
 
-    def list_all_users(self):
-        """List all users in the database
-        """
-        return self.db.list_all_ids()
-
     def exists(self):
         """Check if the user exists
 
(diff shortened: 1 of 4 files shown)


neteler (approved):

mmacata (approved):

src/actinia_core/resources/common/user.py
```diff
@@ -361,11 +361,6 @@ def __str__(self):
 
         return(string)
 
-    def list_all_users(self):
-        """List all users in the database
-        """
-        return self.db.list_all_ids()
-
```
marcjansen, line 368: Are these lines deleted intentionally?

neteler: Yes, it was defined twice:

https://github.com/mundialis/actinia_core/blob/e451a7301409d5c37463b8c2c875d9cb04c69b29/src/actinia_core/resources/common/user.py#L364

https://github.com/mundialis/actinia_core/blob/e451a7301409d5c37463b8c2c875d9cb04c69b29/src/actinia_core/resources/common/user.py#L593

marcjansen: Thanks for the clarification.

PR #117, opened 2020-10-07
Fix docker folder names

mmacata:
title says it all


mmacata:
As done in pair programming, I'll merge now.

Issue #119, opened 2020-10-16
Building and deployment of actinia_core


geography:
Hi all, 
when I try to build and deploy actinia_core in a docker container I get the following error message: 

Starting GRASS GIS...
ERROR: Unable to create new location because the location <utm32n> already exists.
Exiting...
Starting GRASS GIS...
ERROR: Unable to create new location because the location <latlong_wgs84> already exists.
Exiting...
Starting GRASS GIS...
ERROR: Unable to create new location because the location <nc_spm_08> already exists.
Exiting...
/src/start.sh: line 29: actinia-user: not found
/src/start.sh: line 30: actinia-user: not found
Failed to start actinia-user: 127
Starting GRASS GIS...
ERROR: Unable to create new location because the location <utm32n> already exists.
Exiting...
Starting GRASS GIS...
ERROR: Unable to create new location because the location <latlong_wgs84> already exists.
Exiting...
Starting GRASS GIS...
ERROR: Unable to create new location because the location <nc_spm_08> already exists.
Exiting...
/src/start.sh: line 29: actinia-user: not found
/src/start.sh: line 30: actinia-user: not found
Failed to start actinia-user: 127
Starting GRASS GIS...
ERROR: Unable to create new location because the location <utm32n> already exists.
Exiting...
Starting GRASS GIS...
ERROR: Unable to create new location because the location <latlong_wgs84> already exists.
Exiting...
Starting GRASS GIS...
ERROR: Unable to create new location because the location <nc_spm_08> already exists.
Exiting...
/src/start.sh: line 29: actinia-user: not found
/src/start.sh: line 30: actinia-user: not found
Failed to start actinia-user: 127
Starting GRASS GIS...
ERROR: Unable to create new location because the location <utm32n> already exists.
Exiting...
Starting GRASS GIS...
ERROR: Unable to create new location because the location <latlong_wgs84> already exists.
Exiting...
Starting GRASS GIS...
ERROR: Unable to create new location because the location <nc_spm_08> already exists.
Exiting...
/src/start.sh: line 29: actinia-user: not found
/src/start.sh: line 30: actinia-user: not found
Failed to start actinia-user: 127

Someone can give me some hint to try to solve it? 
Thanks in advance for all the work done.
Ema.

neteler:
Please tell us a bit more how you actually build and deploy actinia_core in a docker container.

geography:
Hi Markus, 

I tried this in my windows 10 work environment:

[...]

PR #123, opened 2020-10-27
start.sh: silence error when GRASS GIS DB already created

neteler:
Avoid errors when mounting an existing GRASS GIS DB:

```
sh /src/start.sh
Starting GRASS GIS...
ERROR: Unable to create new location because the location <utm32n> already exists.
Exiting...
Starting GRASS GIS...
ERROR: Unable to create new location because the location <latlong_wgs84> already exists.
Exiting...
Starting GRASS GIS...
ERROR: Unable to create new location because the location <nc_spm_08> already exists.
Exiting...
Created user <actinia-gdi> in group <superadmin>
```

Addresses #119


mmacata:
Great example, how simple things can make daily work much better. Thank you!

mmacata (approved):

PR #124, opened 2020-11-10
Pin versions for proj-using python libs

mmacata:
for ubuntu based dockerimage to avoid error
` ERROR: Minimum supported proj version is 7.2.0, installed version is 5.2.0. For more information see: https://pyproj4.github.io/pyproj/stable/installation.html`
Versions are taken from latest successfully build ubuntu-based Dockerimage.


mmacata:
As discussed, I'll merge now.

PR #136, opened 2021-01-07
repair actinia-user

mmacata:
While `create` and `update` of an existing user was successful, `updating` a non-existing user led to an exception, `creation` of an already existing user ignored the user and `delete` and `show` commands told that the user does not exist.

This PR fixes above:
* Print Error for update of non-existing user
* Print Error for creation of existing user
* Fix delete and show commands

### DIFF
--- a/scripts/actinia-user
+++ b/scripts/actinia-user
@@ -219,7 +219,7 @@ def main():
 
         user = ActiniaUser(user_id=args.user_id, user_group=args.user_group)
 
-        if user.exists() is True:
+        if user.exists() == 1:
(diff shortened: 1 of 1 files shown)


anikaweinmann (approved):

griembauer (approved):

neteler (approved):

marcjansen:
It seems as if the documentation ofthe [exists](https://github.com/mundialis/actinia_core/blob/63bfee64ed6aaab3285039d2b99ca8aa3a6ab285/src/actinia_core/resources/common/user.py#L364-L374) method is wrong, probably something for a follow-up PR?

PR #137, opened 2021-01-08
Saving interim results

anikaweinmann:
After each successfully processed module the temporary mapset is saved as interim result.
The saving consists of either:
1. a `copy`: for the first module
2. an `rsync`:  if the mapset has become smaller in the step
3. an `copy` of the old interim result and than an `rsync`: if the mapset has an equal or greater size 
4. an `rename`: if the sha625sum of the mapset has not changed

The saving of the interim results can be configured in the `actinia.cfg`:
```
...
[MISC]
save_interim_results = True
...
```

The interim resutls are saved to the `grass_resource_dir` in an `interim` folder. Therefore they can be deleted by `DELETE {actinia_url}/api/v1/resource_storage`.

For successful persistent processes the interim results are removed after the mapset merging.


mmacata:
Nice work! Thank you. I have no direct comments to source code, only some general remarks:

- is `rsync` added to the alpine Docker images as well?
- are the interim results for ephemeral processing removed after successful processing as well?
- what happens, if importer/exporter are used and number of steps is a bit fuzzy - the steps will always count up I guess and it won't be a problem? Or might it be helpful to call the interim result folder name by a combination of step + stepid? (We can also decide this later, when we implement using of interim results in processing)

anikaweinmann:
> Nice work! Thank you. I have no direct comments to source code, only some general remarks:
> 
> * is `rsync` added to the alpine Docker images as well?
Yes docker/actinia-core-alpine/Dockerfile_runtime_pkgs line 73
> * are the interim results for ephemeral processing removed after successful processing as well?
No, because there the mapsets are not merged and maybe someone want to have the interims results after the successful processing.
> * what happens, if importer/exporter are used and number of steps is a bit fuzzy - the steps will always count up I guess and it won't be a problem? Or might it be helpful to call the interim result folder name by a combination of step + stepid? (We can also decide this later, when we implement using of interim results in processing)
I do not have checked import and export of data

src/actinia_core/resources/common/config.py
```diff
@@ -128,6 +128,7 @@ def __init__(self):
         self.SECRET_KEY = "This is a very secret key that is used to sign tokens"
         self.DOWNLOAD_CACHE = "/tmp/download_cache"  # The directory to cache downloaded data
         self.DOWNLOAD_CACHE_QUOTA = 100  # The quota of the download cache in Gigibit
+        self.SAVE_INTERIM_RESULTS = False # If set True the interim results (temporary mapset) is saved to
```
neteler, line 131:
```suggestion
        self.SAVE_INTERIM_RESULTS = False # If True the interim results (temporary mapset) are saved
```

src/actinia_core/resources/ephemeral_processing.py
```diff
@@ -1074,6 +1084,101 @@ def _run_module(self, process, poll_time=0.05):
...
+            os.chdir(self.temp_mapset_path)
+            p_cha512sum_curr = os.popen(cha512sum_cmd)
+            cha512sum_curr= p_cha512sum_curr.read().strip()
+            os.chdir(cur_working_dir)
+            del cur_working_dir
+            # compare cha512sums
+            if cha512sum_prev == cha512sum_curr:
+                self.message_logger.info(
+                    "Cha512sum of maspsets are equal; renaming interim result")
+                os.rename(old_dest, dest)
```
neteler, line 1153:
```suggestion
            sha512sum_cmd = "find . -type f -exec sha512sum {} \; | " + \
                            "sort -k 2 | sha512sum"
            # sha512sum of the previous step
            os.chdir(old_dest)
            p_sha512sum_prev = os.popen(sha512sum_cmd)
            sha512sum_prev = p_sha512sum_prev.read().strip()
            # sha512sum of the current step
            os.chdir(self.temp_mapset_path)
            p_sha512sum_curr = os.popen(sha512sum_cmd)
            sha512sum_curr= p_sha512sum_curr.read().strip()
            os.chdir(cur_working_dir)
            del cur_working_dir
            # compare sha512sums
            if sha512sum_prev == sha512sum_curr:
                self.message_logger.info(
                    "Sha512sums of maspsets are equal; renaming interim result")
                os.rename(old_dest, dest)
```

src/actinia_core/resources/ephemeral_processing.py
```diff
@@ -1074,6 +1084,101 @@ def _run_module(self, process, poll_time=0.05):
...
+
+            # rsync or copy folder
+            if size_curr_step > size_prev_step * 0.9:
+                self.message_logger.info("Copy old interim result")
+                shutil.copytree(old_dest, dest)
+            self.message_logger.info(
+                "Rsync mapset %s to interim result" % self.temp_mapset_path)
+            rsync_cmd = [
+                "rsync",
+                "-rzP",
```
neteler, line 1168:
```suggestion
                "--recursive --compress --partial --progress --times",
```

Added `--times` to preserve file modification times (time stamps)
AFAIK the old single dash flags will go away one day and the double dash flags are preferred (and more readable).

src/actinia_core/resources/ephemeral_processing.py
```diff
@@ -1165,7 +1270,17 @@ def _run_executable(self, process, poll_time=0.005):
...
+            raise AsyncProcessError(
+                "Error while running executable <%s>" % process.executable)
+
+        # save interim results
+        if (self.save_interim_results is True
+                and self.temp_mapset_path is not None):
+            self._save_interim_results()
+        elif self.temp_mapset_path is None:
+            self.message_logger.debug(
+                "No temp mapset path set. Because of that no interim results" \
```
neteler, line 1282: Suggestion: please add where to set this (config?):

"No temp mapset path set in ADDME ..."

anikaweinmann: `self.temp_mapset_path` is not set for endpoint like `locations`, `locations/{LOCATION}/mapsets`, ... .
I only added the debug message for us if the interim results are not saved/saved but we expect something other.

neteler: Right - I overlooked that this is debugging only.

PR #143, opened 2021-01-21
Job resumption

anikaweinmann:
This pull request enables to resumpt a job by using a HTTP PUT against the resouce. For the job resumption the interim results have to be configured.

The changes include the following:
- Added an optinal parameters `iteration` and `post_url` to the response model for resuming a resource
- The endpoint **resources/{user_id}/{resource_id}** returns the lates `iteration` if the `resource_id` is like `resource-id_{ID}`. If the `resource_id` is only the `ID` then this endpoint returns all iterations of the resource
- A endpoint **resources/{user_id}/{resource_id}/{iteration}** is added to only get one specific iteration of the resource. The `iteration` has to be a integer.

- Added **HTTP PUT** for **resources/{user_id}/{resource_id}** for resumption a resource.
  This enques the job with the `resource_id` and add an `iteration` and `post_url` for it. By validation the process chain is trimmed to only the process steps which has not runned before.
(Tests including importer, exporter and stdout are created for the job resumption)


src/actinia_core/endpoints.py
```diff
@@ -201,13 +201,25 @@ def create_core_endpoints():
...
     flask_api.add_resource(APILogResource, '/api_log/<string:user_id>')
+
     # Resource management
+    """
+    The endpoint '/resources/<string:user_id>/<string:resource_id>' has to
+    different answerd depending on the resource_id. If the resoucre_id starts
+    with 'resouce-id_' the latest iteration of the resouce is given back.
+    If the resocue_id is only the id than all iterations of the resource are
+    given in the response.
+    """
```
mmacata, line 212: `has to different answerd` = has a different answer?
Also some typos in this block: `resoucre_id`, `'resouce-id_'`, `resouce`, `resocue_id` and `is only the id than` -> then

src/actinia_core/resources/common/config.py
```diff
@@ -172,9 +172,12 @@ def __init__(self):
         # The directory to store temporary files
         self.TMP_WORKDIR = "%s/actinia/workspace/tmp" % home
         self.SECRET_KEY = "This is a very secret key that is used to sign tokens"
-        self.DOWNLOAD_CACHE = "/tmp/download_cache"  # The directory to cache downloaded data
-        self.DOWNLOAD_CACHE_QUOTA = 100  # The quota of the download cache in Gigibit
-        self.SAVE_INTERIM_RESULTS = False # If True the interim results (temporary mapset) are saved
+        # The directory to cache downloaded data
+        self.DOWNLOAD_CACHE = "/tmp/download_cache"
+        # The quota of the download cache in Gigibit
```
marcjansen: Gigibit?

src/actinia_core/resources/common/process_queue.py
```diff
@@ -258,11 +258,12 @@ def _send_resource_update(self, status, message, response_data=None):
...
-            self.resource_logger.commit(user_id=self.user_id,
-                                        resource_id=self.resource_id,
-                                        itertion=self.itertion,
-                                        document=document,
-                                        expiration=self.config.REDIS_RESOURCE_EXPIRE_TIME)
+            self.resource_logger.commit(
+                user_id=self.user_id,
+                resource_id=self.resource_id,
+                itertion=self.itertion,
+                document=document,
```
marcjansen, line 270: itertion = iteration?

src/actinia_core/resources/ephemeral_processing.py
```diff
@@ -1455,7 +1457,7 @@ def _run_executable(self, process, poll_time=0.005):
             self._save_interim_results()
         elif self.temp_mapset_path is None:
             self.message_logger.debug(
-                "No temp mapset path set. Because of that no interim results" \
+                "No temp mapset path set. Because of that no interim results"
```
marcjansen: This looks odd

marcjansen (commented):
Just some tiny observations.

src/actinia_core/resources/common/config.py
neteler: I suppose it should indeed read Gig**a**bit (as being based on 1024 and not 1000), see:

https://github.com/mundialis/actinia_core/blob/95132551ed785ce498ebe90151363253c1fd9618/src/actinia_core/resources/download_cache_management.py#L144

mmacata (commented):
Great work! I have only one minor comment. In general I approve.

mmacata (approved):

PR #148, opened 2021-02-25
docker: update module-plugin

mmacata:


mmacata:
As only minor in alpine Dockerfile, I will merge directly

PR #152, opened 2021-03-05
Fix several misspelled words

marcjansen:
These were found with [`codespell`](https://github.com/codespell-project/codespell).

I was unsure of what to do with `Gigibit` in the following two cases:

* https://github.com/mundialis/actinia_core/blob/master/src/actinia_core/resources/common/config.py#L86
* https://github.com/mundialis/actinia_core/blob/master/src/actinia_core/resources/common/config.py#L130

Are we talking about `Gigabyte` or `Gigabit`? Can surely be fixed in a follow-up request.

Please review.


hwbllmnn (approved):

marcjansen:
Thanks @hwbllmnn, I'd favor another review from @mundialis-dev

neteler (approved):

marcjansen:
Thanks friends.

PR #153, opened 2021-03-05
Don't use mutable default arguments when avoidable

marcjansen:
Having mutable default arguments is sometimes [considered a code smell / common pitfall](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/), and [some suggest](https://towardsdatascience.com/5-python-code-smells-you-should-be-wary-of-c48cc0eb9d8b) to avoid them whenever possible.

This PR suggests to replace some occurences of lists as default arguments with an appropriate assignment, if the argument wasn't given (i.e. `None`).

Please review.


hwbllmnn (approved):

marcjansen:
Thanks @hwbllmnn, I'd favor another review from a mundialis dev.

marcjansen:
Thanks everybody for the review!

PR #155, opened 2021-03-10
Use non-root user for prod docker example

mmacata:
This PR
* lets actinia run as non-root user in prod docker example
* syncs prod docker setup to dev
* removes duplicated lines which are already included in parent image


anikaweinmann (approved):
Thank you please merge

mmacata:
As this is only an example, I will merge now.

Issue #158, opened 2021-03-15
New Landsat directory scheme on Google GCS: scene update needed


neteler:
At time the test `test_landsat_dowunload_commads()` fails (besides the fact, that it should be renamed to `test_landsat_download_commands()`, leading to blocked docker image creation:

```
======================================================================
ERROR: test_landsat_download_commands (__main__.GeoDataDownloadImportSupportTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_geodata_import.py", line 199, in test_landsat_download_commands
    pl, import_file_info = lp.get_download_process_list()
  File "/usr/lib/python3.8/site-packages/actinia_core/resources/common/geodata_download_importer.py", line 149, in get_download_process_list
    self._setup()
  File "/usr/lib/python3.8/site-packages/actinia_core/resources/common/landsat_processing_library.py", line 178, in _setup
    for suffix in SCENE_SUFFIXES[self.landsat_sensor_id]:
KeyError: 'LC00'
```

Reason:
https://github.com/mundialis/actinia_core/blob/e66b6b5d2627e44624711b0a4c4a8e8f8881ce47/tests/test_geodata_import.py#L190

The (old) test dataset `scene_id="LT41970251990147XXX03"` no longer exists  (https://storage.googleapis.com/gcp-public-data-landsat/LT04/PRE/006/016/LT40060161989006XXX02/LT40060161989006XXX02_MTL.txt).

An existing scene is for example ` scene_id="LC08_L1GT_001004_20130910_20170502_01_T2"` (
https://storage.googleapis.com/gcp-public-data-landsat/LC08/01/001/004/LC08_L1GT_001004_20130910_20170502_01_T2/LC08_L1GT_001004_20130910_20170502_01_T2_B1.TIF).
 
The Landsat tests requires an update of the `extract_sensor_id_from_scene_id()` function to properly recognize the new scene directory structure. It is located here:
https://github.com/mundialis/actinia_core/blob/0d132114fe6a6007b25b06d8f485f784598e5470/src/actinia_core/resources/common/landsat_processing_library.py#L68

This is then input to `scene_id_to_google_url()` which constructs the download URL:
https://github.com/mundialis/actinia_core/blob/0d132114fe6a6007b25b06d8f485f784598e5470/src/actinia_core/resources/common/landsat_processing_library.py#L82

At time actinia docker image creation is failing due to this issue.

neteler:
[...]

PR #159, opened 2021-03-17
fix landsat test scene_id and scene_id parser

anikaweinmann:
This MR fixes the landsat test reported in https://github.com/mundialis/actinia_core/issues/158


neteler (approved):
Great, thanks a lot!

Issue #144, opened 2021-01-22
Allow configuration via environment variables


mmacata:
For deployment with kubernetes it would be helpful to be able to configure certain values via environment variable instead of one config file containing all values. This is especially usefull for the redis passwort as it can be set as secret from redis chart.

mmacata:
Fixed in https://github.com/mundialis/actinia_core/pull/164

PR #164, opened 2021-03-31
Allow redis config via env var

mmacata:
See #144

### DIFF
--- a/src/actinia_core/resources/common/config.py
+++ b/src/actinia_core/resources/common/config.py
@@ -478,5 +478,13 @@ class Configuration(object):
                     self.S3_AWS_RESOURCE_BUCKET = config.get(
                         "AWS_S3", "S3_AWS_RESOURCE_BUCKET")
 
+        # Overwrite values with environment variables if exist:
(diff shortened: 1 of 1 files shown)


griembauer (approved):

src/actinia_core/resources/common/config.py
```diff
@@ -478,5 +478,13 @@ def read(self, path=DEFAULT_CONFIG_PATH):
                     self.S3_AWS_RESOURCE_BUCKET = config.get(
                         "AWS_S3", "S3_AWS_RESOURCE_BUCKET")
 
+        # Overwrite values with environment variables if exist:
+        if os.environ.get('REDIS_SERVER_URL'):
```
anikaweinmann, line 482: Maybe a warning `if config.has_option("REDIS", "REDIS_SERVER_URL")` is also given would be nice that says that the env values are taken. But that is not so important to me otherwise, you may merge. Thanks for the change :)

anikaweinmann (approved):

anikaweinmann (approved):

anikaweinmann:
Thanks. Please merge

PR #173, opened 2021-04-13
Update config.py

ahalota:
Create the config file in case it doesn't exist yet (ran into this issue on initial install).

Haven't verified this code change but it's pretty straightforward.


marcjansen:
Thanks for your continued feedback and work here, @ahalota. This is much appreciated. I think we'll give this a proper review later today.

anikaweinmann (approved):
Thank you very much. I tested it. Locally it worked very well. I will merge it.

mmacata (approved):
Thanks!

PR #178, opened 2021-04-16
raise AsyncProcessError if mapset-lock to be checked does not exist

griembauer:
When the `api/v1/locations/<location>/mapsets/<mapsets>/lock` endpoint is used to check whether a mapset is locked, the process is successful and the actinia response contains
```
[...]  
"message": "Mapset lock state: False",
"process_chain_list": [],
"process_log": [],
"process_results": false,
[...]
```
even if the mapset does not exist. With this PR, an `AsyncProcessError` is raised if the combination of location and mapset to be checked for a lock does not exist. In case *not* raising an error is a desired behaviour, there might be another option to let the user know the mapset does not exist, e.g. by providing a different `self.finish_message`.

### DIFF
--- a/src/actinia_core/resources/mapset_management.py
+++ b/src/actinia_core/resources/mapset_management.py
@@ -649,8 +649,16 @@ class PersistentGetMapsetLock(PersistentProcessing):
 
     def _execute(self):
         self._setup()
-        self.module_results = self.lock_interface.get(self.target_mapset_lock_id)
-        self.finish_message = "Mapset lock state: %s" % str(self.module_results)
+        exists = self._check_mapset(self.mapset_name)
+        if exists is False:
+            raise AsyncProcessError(
+                "Invalid mapset <%s> in location <%s>"
+                % (self.mapset_name, self.location_name))
+        else:
+            self.module_results = self.lock_interface.get(
+                self.target_mapset_lock_id)
+            self.finish_message = "Mapset lock state: %s" % str(
+                self.module_results)
 
 
 def lock_mapset(*args):
--- a/src/actinia_core/resources/mapset_management.py
+++ b/src/actinia_core/resources/mapset_management.py
@@ -649,10 +649,11 @@ class PersistentGetMapsetLock(PersistentProcessing):
 
     def _execute(self):
         self._setup()
-        exists = self._check_mapset(self.mapset_name)
-        if exists is False:
+        self._check_target_mapset_exists()
+        if self.target_mapset_exists is False:
             raise AsyncProcessError(
-                "Invalid mapset <%s> in location <%s>"
+                ("Unable to get lock status of mapset <%s> in location <%s>:"
+                 " Mapset does not exist")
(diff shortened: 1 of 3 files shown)


anikaweinmann (approved):

anikaweinmann:
@griembauer can you please still look at the error in the test? ERROR in: tests/test_mapset_management.py::MapsetTestCase::test_mapset_creation_and_locking
--> /src/actinia_core/tests/test_mapset_management.py:197: AssertionError: 400 != 200 : HTML status code is wrong 400

griembauer:
> @griembauer can you please still look at the error in the test? ERROR in: tests/test_mapset_management.py::MapsetTestCase::test_mapset_creation_and_locking
> --> /src/actinia_core/tests/test_mapset_management.py:197: AssertionError: 400 != 200 : HTML status code is wrong 400

Thanks for pointing this out, the status code changed to 400 so this is now also adapted in the test. However, there is one open question: The last test of `/src/actinia_core/tests/test_mapset_management.py` is a deletion of a lock of a non-existing mapset, which expects a status code 200. Following the logic of this PR, there should also be a status code 400 and an exception raised that this mapset does not exist. Unless of course, it is desired to **not** throw an Exception.

mmacata:
I like the idea that a user will be informed that the mapset does not exist instead of assuming it is not locked.
How does the response look like? I would't mind the `AsyncProcessError` if the user is informed accordingly - and maybe HTTP status code `404` - Not Found might fit a bit better.
I also agree on the changes to `/src/actinia_core/tests/test_mapset_management.py` - the DELETE method should return a similar response to the HTTP GET method with code 400 / 404 as well.

griembauer:
The methods are now adapted to match existing methods in `mapset_management.py` and `persistent_processing.py`. Now an error is also raised if a non-existing mapset is unlocked. The error messages are:

- trying to get the lock status of a non-existing mapset: `"AsyncProcessError:  Unable to get lock status of mapset <mapset> in location <location>: Mapset does not exist"`
- trying to unlock a non-existing mapset: `AsyncProcessError:  Unable to unlock mapset <mapset> in location <location>: Mapset does not exist`

Trying to lock a non-existing mapset already fais with a similar error message before this PR.
However I am unsure how to manipulate the http code, @mmacata can you point me to where this could be implemented?

mmacata:
Great!
For the status code - I didn't know this endpoint enqueues the job as well. Then it is much harder and 400 is fine for me. It most likely comes from [here](https://github.com/mundialis/actinia_core/blob/master/src/actinia_core/resources/ephemeral_processing.py#L502) and is - without custom code - not easy to change. It could be overwritten [here](https://github.com/mundialis/actinia_core/blob/83db2e48e1708a3137060f8e138ecf694fb0fa88/src/actinia_core/resources/mapset_management.py#L541), but as an exception is raised, the code might propably not execute up to this point.

So all fine from my point of view!

mmacata (approved):

griembauer:
Thanks for the additional info and the review! I will merge now

PR #183, opened 2021-04-30
Introducing changelog

mmacata:
Title says it all.

From now on it would be great for all releases, if we mention all done PRs and commits. Therefore it would be very helpfull to use the "squash commits" merge option. In the release notes we can sort them by the suggested categories.

I went through all old releases from last one to v0.99.0 to regain this information but did no sorting.
To edit older releases in the changelog, it would be good to do this directly at the release and then recreate that release information via the command at the end of the file.


griembauer (approved):
Thank you, very helpful!

anikaweinmann (approved):
Thank you

neteler (approved):
Excellent, thanks!

(Importantly, actinia committers need to write understandable commit titles :-) )

PR #200, opened 2021-05-26
Get all mapsets locks

juleshaas:
This PR adds a new endpoint `<actinia-base-url>/locks` to return a list of all currently locked mapsets


src/actinia_core/resources/mapset_management.py
```diff
@@ -716,3 +720,49 @@ def _execute(self):
...
+                'schema': LockedMapsetListResponseModel
+            },
+            '500': {
+                'description': 'The error message and a detailed error log',
+                'schema': SimpleResponseModel
+            }
+        }
+    })
+    def get(self):
+        redis_interface = RedisBaseInterface()
```
mmacata, line 747: You can use directly the [RedisLockingInterface](https://github.com/mundialis/actinia_core/blob/main/src/actinia_core/common/redis_lock.py#L39) here. As far as I can see, it is also used by the other classes which check for existing mapset locks ([here](https://github.com/mundialis/actinia_core/blob/main/src/actinia_core/resources/mapset_management.py#L659)).

src/actinia_core/resources/mapset_management.py
```diff
@@ -716,3 +720,49 @@ def _execute(self):
...
+        kwargs = dict()
+        kwargs["host"] = global_config.REDIS_SERVER_URL
+        kwargs["port"] = global_config.REDIS_SERVER_PORT
+        if global_config.REDIS_SERVER_PW and global_config.REDIS_SERVER_PW is not None:
+            kwargs["password"] = global_config.REDIS_SERVER_PW
+        redis_interface.connect(**kwargs)
```
mmacata, line 753: It might be good to disconnect after receiving the needed information from redis. Another possibility would be to use a `with` statement. I am not sure if this is really needed. Other pieces of code here don't explicitly disconnect but reuse existing connections to redis. The ephemeral processing connects as well without disconnecting but as it is called differently, it might behave differently. I am not sure but I guess it won't hurt either.

src/actinia_core/endpoints.py
```diff
@@ -101,6 +102,9 @@ def create_core_endpoints():
     flask_api.add_resource(
         MapsetLockManagementResource,
         '/locations/<string:location_name>/mapsets/<string:mapset_name>/lock')
+    flask_api.add_resource(
+        MapsetLockManagementResourceAdmin, '/locations/locks')
```
mmacata, line 106: Usually in actinia, after `/locations/` the adressed location name is followed. As this makes no sense in this case, I would prefer a route without `locations` in it at all. I cannot think of a good name yet but will try again tommorow :upside_down_face:

mmacata (changes requested):

src/actinia_core/endpoints.py
mmacata: Just for documentation: We agreed upon a new Resource `/mapsets` independant of a location. For now, it will only be accessible by HTTP GET parameter filter and empty response without filter.

The class will be moved away from `mapset_management.py` into a new one.

mmacata (approved):
Very cool, thank you! Please merge!

marcjansen:
:100: :100: 

Great contribution! Thanks

PR #202, opened 2021-05-26
Docker Readme

juleshaas:
This MR adds information to the "Local dev-setup with docker" part in the Docker Readme.


mmacata:
Thanks for this enhancement! It looks like this branch was created from #200 as the changes are included here as well. Either please reopen directly from main branch, or we wait until the other PR is ready and merged.

juleshaas:
I would wait for the #200 then, as it is only a minor change.

Issue #190, opened 2021-05-20
Introduce subfolders / restructure modules


mmacata:
title says it all...

mmacata:
The folder structure to start with was agreed upon:
```
.
├── api
│   ├── topic1
│   └── topic2
├── common
│   ├── topic1
│   └── topic3
└── model
    ├── topic1
    └── topic2
```
while topic relates to colored boxes of this UML draft

![actinia](https://user-images.githubusercontent.com/8308401/119654762-13335580-be29-11eb-9c87-6aa8a11d3b74.png)

mmacata:
Status is:
Code was sorted into `rest`, `core` and `models` folders
* modules from `actinia_core.resources.common` changed to `actinia_core.core`
  * but modules which are used by actinia plugins changed to `actinia_core.core.common`
* modules from `actinia_core.resources` changed to `actinia_core.rest`
* response_models moved to `actinia_core.models.response_models`
* models were moved out of some modules into separate model modules into `actinia_core.models.openapi`

PR #207, opened 2021-05-27
Rename resources folder folder to rest + use absolute imports

mmacata:
Part of #190

This PR moves all modules from actinia_core.resources to actinia_core.rest. For all imports which needed to be changed, absolute imports were used instead of relative ones.


src/actinia_core/rest/ephemeral_processing.py
```diff
@@ -535,9 +536,42 @@ def _send_to_database(self, document, final=False):
         # Call the webhook after the final result was send to the database
         try:
             if final is True and self.webhook_finished is not None:
-                self._post_to_webhook(document, 'finished')
+                self.message_logger.info(
```
anikaweinmann, line 539: It looks like here is no main branch merged? Or an old version?

anikaweinmann (approved):

Issue #211, opened 2021-05-28
Check install steps


mmacata:
Follow the installation steps from https://actinia.mundialis.de/tutorial/installation.html and document deprecated information in this ticket or directly in the documentation.
Attention: the documentation changed from rst files to md: https://github.com/mundialis/actinia_core/blob/main/docs/docs/installation.md

neteler:
(please feel free to directly edit the new `installation.md` file and propose changes as PR, thanks)

sabman:
@mmacata I'm taking a look at this now for the following distro
```
Linux li1502-8 4.19.0-17-amd64 #1 SMP Debian 4.19.194-3 (2021-07-18) x86_64 GNU/Linux
```

sabman:
I'm running into issues at

sabman:
So I got up to https://github.com/mundialis/actinia_core/blob/main/docs/docs/installation.md?plain=1#L80-L85 after following dependencies where met:

```
apt install zlib1g-dev unzip build-essentials libbz2-dev \
  libzstd-dev gdal-bin libgdal-dev libfftw3-dev libcairo2-dev
```

However the next step is causing the following error. 

```
root@li1502-8:/tmp/grass_master/temporal# cd t.rast.sample/
root@li1502-8:/tmp/grass_master/temporal/t.rast.sample# make
[...]

PR #232, opened 2021-07-28
Fix "check service" URL in install instructions

ashiklom:
The `/api/v1` is required -- otherwise, you get a cryptic "404 Not Found" HTML error.


ashiklom:
Related to #211.

mmacata (approved):
Thanks for that! I will merge.

Issue #192, opened 2021-05-21
Add "tgis" support in actinia for persistent processing and export


neteler:
At time the directory `tgis` is not yet copied:

https://github.com/mundialis/actinia_core/blob/e1510fce08f8d40d5fc048dc7fd81e920ed5ffc3/src/actinia_core/resources/persistent_processing.py#L477

In addition, the support needs to be added also in some other places:
- update of mapset name in tgis sqlite.db in persistent part
- update mapset name in tgis sqlite.db when creating temporary DB
- add to exporter (separate from raster export)
    - e.g. netCDF
    - in case, augment also https://grass.osgeo.org/grass-devel/manuals/t.rast.export.html and https://grass.osgeo.org/grass-devel/manuals/t.vect.export.html

For inspiration, see
https://redocly.github.io/redoc/?url=https://actinia.mundialis.de/latest/swagger.json#tag/STRDS-Management

metzm:
Updating the mapset names in the tgis db is not easy: you have to check the current tgis db connection as defined in the `<mapset>/VAR` file, then use custom SQL commands to update the correct table and most probably some views need to be recreated.

neteler:
Here a dirty hack to change mapset names in `tgis/sqlite.db`:

1. dump SQLite DB, change mapset names on the fly and save to new dump file
2. restore new dump file as a new `tgis/sqlite.db`:

```
/actinia_core/grassdb/geoharmonizer_aq/PERMANENT/tgis # sqlite3 sqlite.db .dump | sed 's+PERMANENT+mymapset+g' > dumpfile
/actinia_core/grassdb/geoharmonizer_aq/PERMANENT/tgis # cat dumpfile | sqlite3 my_database.sqlite

```

anikaweinmann:
Parts are done in #233, but `t.rast.extract` does not work yet or creating a STRDS from rasters from a previous PC (`t.create` + `t.register`) does not work either. For this I have created a new bug ticket in GRASS GIS: https://github.com/OSGeo/grass/issues/1791

metzm:
A first step for TGIS support has been done in GRASS with https://github.com/OSGeo/grass/pull/1924 to register maps from a different mapset in a stds in the current mapset.

metzm:
TGIS has been enhanced in GRASS 8.0 to work with maps that are not in the current mapset.

Waiting for a new deployment of actinia for testing.

anikaweinmann:
[...]

PR #233, opened 2021-07-30
Integrate tgis in mapset merge

anikaweinmann:
This changes are part of #192 and integrate the tgis folder in the mapset merge.

Created with @griembauer.

Added to this PR: If the mapsets allready exists and have a tgis sqlite.db file then this is merged in the temporary sqlite.db and then copied to the target mapset.

Test are also added:
1. create a STRDS modis
2. create a STRDS modis and in a second PC list the modis STRDS
3. create a STRDS modis and in a second PC create a other STRDS modis2
4. create a STRDS modis and in a second PC create a other STRDS modis2 and a modis STRDS with other input as in the first PC

### DIFF
--- a/src/actinia_core/rest/persistent_processing.py
+++ b/src/actinia_core/rest/persistent_processing.py
@@ -25,12 +25,13 @@
 Asynchronous computation in specific temporary generated and then copied
 or original mapsets
 """
-import pickle
+import fileinput
 import os
+import pickle
 import shutil
+import sqlite3
 import subprocess
 from copy import deepcopy
-import fileinput
 from flask import jsonify, make_response
 from flask_restful_swagger_2 import swagger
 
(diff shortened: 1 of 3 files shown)


mmacata (approved):

mmacata:
Cool new feature! Thanks for your work!

src/actinia_core/rest/persistent_processing.py
```diff
@@ -478,22 +479,152 @@ def _lock_target_mapset(self):
...
+                                   target_mapset, target_tgis_db):
+        """Replaces the mapset name in the tgis sqlite.db
+
+        Args:
+            tgis_path(str): path of the tgis folder in the source mapset
+            source_mapset(str): name of source mapset
+            target_mapset(str): name of target mapset
+
+        Raises:
+            This method will raise an AsyncProcessError if a group has no REF file
```
griembauer, line 585: I think this is just copied from above and can be removed

src/actinia_core/rest/persistent_processing.py
```diff
@@ -502,16 +633,15 @@ def _merge_mapset_into_target(self, source_mapset, target_mapset):
...
-                                print(line.replace(
-                                    source_mapset, target_mapset), end='')
-                        else:
-                            raise AsyncProcessError("group %s has no REF file"
-                                                    % (group_dir))
+                    self._change_mapsetname_in_group(
+                        source_path, source_mapset, target_mapset)
+                if directory == "tgis":
+                    target_tgis_db = None
+                    if os.path.isdir(os.path.join(target_path, 'tgis')):
```
griembauer, line 640: It might be better to check for the actual existence of `target_path/tgis/sqlite.db` - I don't know if there can be a case that the folder `tgis` exists, but not the `tgis/sqlite.db`

griembauer (approved):
Thanks, great feature! Just two very small (and optional) comments

src/actinia_core/rest/persistent_processing.py
mmacata: I don't think so, only if the file is manually deleted. I tested it locally - I had one STRDS and removed it. Folder `tgis` and file `sqlite.db`  both still exist, but table `strds_base` is empty.

src/actinia_core/rest/persistent_processing.py
griembauer: Was updated in https://github.com/mundialis/actinia_core/pull/233/commits/7f454e717bcb71019144b5d36e6cfbd1639832e9

mmacata (approved):

Issue #223, opened 2021-06-20
Are all of the installed GRASS GIS AddOns in Dockerfiles necessary for actinia core?


pesekon2:
In the [alpine](https://github.com/mundialis/actinia_core/blob/main/docker/actinia-core-alpine/Dockerfile#L50) and [Ubuntu](https://github.com/mundialis/actinia_core/blob/main/docker/actinia-core-ubuntu/Dockerfile#L76) Dockerfiles and therefore also in images uploaded to dockerhub, all GRASS GIS AddOns specified [here](https://github.com/mundialis/actinia_core/blob/main/docker/grass_addons_list.csv) are installed.

I believe that most of them are not necessary for the official core images and although they do not take a lot of space, they should be removed.

neteler:
We use the official core images for the demo server (https://actinia.mundialis.de/).

Definitely needed are
- importer | https://github.com/mundialis/importer
- exporter | https://github.com/mundialis/exporter

The others are debatable. What exactly is the issue with including them?

pesekon2:
Well, there is no real issue with them, they do not conflict with anything and the increase in the size of the image is not noticable at all. 

It's just my expectation that in a core image, only core features would be included; anything that is unnecessary and not connected to actinia should be a task of anyone building his/her/its project on top of that - for example having a demo image that is used at mundialis and is just the core image with some additional AddOns. But maybe this is just my approach and not a general one - feel free to close the issue in such case.

neteler:
It's a good suggestion, we'll discuss it internally (asap).

pesekon2:
I could make a PR in the case you would find that reasonable, but probably you know better what are all the things that would not be necessary for the core image (the locations download too, I believe).

mmacata:
In general I agree with this approach to only have plain core functionality. The case with these addons is that we use them with many actinia installations so we decided to have only one place for them. For some installations we don't have a build stage at all but configure everything only during deployment. This way, we would need an additional official actinia docker image. For complexity reasons I would vote here to keep them inside the Dockerfile. If e.g. something in GRASS GIS changes, currently we wait for grass-py3-pdal to build and then rebuild actinia. Depending on where it is installed, we even need to make an actinia release and rebuild. When we remove these addons, we would even need to wait for another build before we can deploy. This build chain is not ideal yet and I would prefer to not increase the complexity further.

pesekon2:
Well, okay, it's your chain in the end. Closing the issue then.

PR #235, opened 2021-08-11
add openeo addons

mmacata:
This PR adds more openeo GRASS GIS addons to the default actinia build.


metzm (approved):
I'm not fully happy with including all these addons here in core, but an alternative solution seems to be overly complex.

mmacata:
You are not the first one - https://github.com/mundialis/actinia_core/issues/223. As they grow more and more, I agree. We might find a different solution in the future, but as no separate build step is needed in some installations, I see currently no fully automated way. Also https://github.com/mundialis/helm-charts/tree/main/charts/actinia can right now simply be used without extra build.

PR #236, opened 2021-08-18
Add support for STRDS export

mmacata:
This PR initiates export for STRDS. Here, `t.rast.export` was used which does not support netCDF or similar formats but writes out separate GeoTIFFs and packs them. The output structure of the archive can be seen in the [GRASS GIS documentation](https://grass.osgeo.org/grass79/manuals/t.rast.export.html).

Part of https://github.com/mundialis/actinia_core/issues/192

It was tested locally with `/api/v1/locations/nc_spm_08/processing_async_export` endpoint with modis_lst mapset available. The process chain used was
```
{
    "list": [
      {
        "id": "t_rast_extract_93486",
        "inputs": [
          {
            "param": "input",
            "value": "LST_Day_monthly@modis_lst"
          },
          {
            "param": "where",
            "value": "start_time >= '2016-05-01' AND start_time <= '2016-05-30'"
          },
          {
            "param": "output",
            "value": "lst_test"
          },
          {
            "param": "expression",
            "value": "1.0 * LST_Day_monthly@modis_lst"
          },
          {
            "param": "basename",
            "value": "lst_test"
          }
        ],
        "module": "t.rast.extract"
      },
      {
        "id": "save_result_333549",
        "module": "exporter",
        "outputs": [
          {
            "export": {
              "format": "GTiff",
              "type": "strds"
            },
            "param": "map",
            "value": "lst_test"
          }
        ]
      }
    ],
    "version": "1"
}
```


src/actinia_core/rest/ephemeral_processing_with_export.py
```diff
@@ -294,6 +294,51 @@ def _export_raster(self, raster_name,
...
+
+        module_name = "t.rast.export"
+        args = [
+            "input=%s" % strds_name,
+            "format=%s" % format,
+            "output=%s" % output_path,
+            "compression=%s" % "bzip2"
+        ]
+        # optimized for GTiff
+        create_opts = "createopt=BIGTIFF=YES,COMPRESS=LZW,TILED=YES"
```
griembauer, line 327: Might it be necessary to use single quotes?
`create_opts = "createopt='BIGTIFF=YES,COMPRESS=LZW,TILED=YES'"`

griembauer (commented):
Thanks a lot, looks good to me, just one small question

src/actinia_core/rest/ephemeral_processing_with_export.py
mmacata: in the actinia output it looks like this:
```
   {
      "executable": "t.rast.export",
      "id": "exporter_strds_lst_test",
      "mapset_size": 731903,
      "parameter": [
        "input=lst_test",
        "format=GTiff",
        "output=/actinia_core/workspace/temp_db/gisdbase_0d884d0430814e97b3417abf9163f91b/.tmp/lst_test.tar.bzip2",
        "compression=bzip2",
        "createopt=BIGTIFF=YES,COMPRESS=LZW,TILED=YES"
      ],
 ```
 
for existing raster export it looks syntactically the same:
```
    {
      "executable": "r.out.gdal",
      "id": "exporter_raster_lst_test_extract_00001",
      "mapset_size": 731909,
      "parameter": [
        "-fmt",
        "input=lst_test_extract_00001",
        "format=GTiff",
        "output=/actinia_core/workspace/temp_db/gisdbase_d[...]3/.tmp/lst_test_extract_00001.tif",
        "overviews=5",
        "createopt=BIGTIFF=YES,COMPRESS=LZW,TILED=YES"
      ],
```
gdalinfo tells me that it has LZW compression `COMPRESSION=LZW`

So I guess it is fine that way?

Comment (griembauer) [reply src/actinia_core/rest/ephemeral_processing_with_export.py:327]: I agree, if it is the same for an existing (working) raster export, it should be fine.

**Review by griembauer — APPROVED** [review]

File: src/actinia_core/rest/ephemeral_processing_with_export.py [hunk 2]
```diff
@@ -294,6 +294,51 @@ def _export_raster(self, raster_name,
...
+
+        Args:
+            strds_name (str): The name of the strds layer
+            format (str): GTiff (only option)
+
+        Returns:
+            tuple: A tuple (file_name, output_path)
+
+        """
+        suffix = ".tar.bzip2"
```
metzm, line 311: bzip2 works really nice with text and integer data, but not with already compressed data, and here the tifs are already compressed with COMPRESS=LZW. Generally, compressors struggle to compress already compressed data, there is usually little to no size reduction, and it takes quite long. Therefore I would recommend `t.rast.export compression=no`

metzm: Some quick test shows that both bzip2 and gzip do reduce file size of a tar with LZW compressed GeoTIFFs of Float64. Example is the process chain [above](https://github.com/mundialis/actinia_core/pull/236#issue-715097084 with `compression=no`, then using bzip2 and gzip on the tar file:
```
 122880 Aug 18 19:42 lst_test.tar
  77758 Aug 18 19:42 lst_test.tar.bz2
  76299 Aug 18 19:42 lst_test.tar.gz
```
Thus I recommend `compression=gzip`. gzip might in theory not compress as heavily as bzip2, but if further compression fails to reduce file size, it is faster than bzip2, thus it seems to be a reasonable compromise.

mmacata: Thanks for the test! I changed it to suggested `gzip`.

metzm, line 311: One last thing:
```suggestion
        suffix = ".tar.gz"
```

metzm (changes requested):

metzm (approved):

mmacata:
Thanks for the reviews, I will merge now!

PR #238, opened 2021-08-20
Prevent duplicate workflow run for release

mmacata:
Currently, GHA for docker builds is triggered for releases when they are created and published. This makes sense, when pre-releases are created but not in our case. This PR reduces the trigger to the creation of a release.


mmacata:
As only minor, I will merge directly.

PR #252, opened 2021-10-02
Make the version output more flexible

### LINKED ISSUE #175: Add a project version to the version endpoint
It would be nice to have a possibility to specify a project version which can be seen in the version endpoint. 
Maybe the version can been set via an environment variable and if it is set, it will be displayed by requesting e.g. https://actinia.mundialis.de/api/v1/version. 
It cood look like this:
```
{
      "grass_version": {...},
      "plugin_versions": {...},
      "plugins": "actinia_statistic_plugin,actinia_satellite_plugin,actinia_gdi",
      "project_version": "1.0.0",
      "python_version": "3.8.5 (default, Jul 20 2020, 23:11:29) - [GCC 9.3.0]",
      "version": "0.99.19"
}
```

marcjansen:
This allows users to configure what will be reported when requesting the `/version`-endpoint.

In case a user has the environment variable `ACTINIA_ADDITIONAL_VERSION_INFO` set  its content will be parsed as key-value-pairs (key and value seperated by colon `:`, pairs seperated by  pipe `|`). 

The content of the environment variable `ACTINIA_RUNNING_SINCE` will also be reported, if it is set. Contrary to what is written in #175 I named the environment variable `…RUNNING…` instead of `…DEPLOYED…`, as I think the most sensible way to provide this in real world settings is to set the enviroinment variable in a startup script or such.

We should probably document this new behaviour and also tell users not put sensitive data inside of this environment variables

I failed at creating unit tests for the new methods, if I get a pointer how to do this, that'd be much appreciated.

Please review

Fixes #175

### DIFF
--- a/src/actinia_core/version.py
+++ b/src/actinia_core/version.py
@@ -32,6 +32,8 @@ __copyright__ = "Copyright 2016-2018, Sören Gebbert and mundialis GmbH & Co. KG
 __maintainer__ = "mundialis"
 
 from flask import make_response, jsonify
+import os
+import re
 import importlib
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


anikaweinmann (approved):
Thank you :)

marcjansen:
Thanks for the review.

PR #256, opened 2021-10-04
Use recommended way to create temporary files

marcjansen:
The function `mktemp` has been deprecated since Python 2.3 and the replacement `NamedTemporaryFile` should be used to ensure that no other process accidentially writes to the file we expect to be unique for us.

This addresses a common weakness "CWE-377: Insecure Temporary File" see e.g. https://cwe.mitre.org/data/definitions/377.html for more information.

This is not ready for review yet, as there are multiple instances where we use `mktemp`. Once I see how the build reacts to this, the others will follow.


anikaweinmann:
@marcjansen I change some things and now the tests are working.
The problem was, that the `overwrite` was not correct in the PC: `ERROR: r.colors.out: Sorry, <overwrite> is not a valid parameter`. Actinia sets in GRASS GIS the overwrite by default auf True, so that you can not unset it in a process chain: https://github.com/mundialis/actinia_core/blob/main/src/actinia_core/core/grass_init.py#L209

marcjansen:
Thanks for the great help @anikaweinmann, much appreciated! Also thanks for the review, @griembauer and @anikaweinmann. I'll merge now.

PR #257, opened 2021-10-04
allow endpoints with and without trailing slash

juleshaas:
This PR adds the functionality to use endpoints both with and without trailing backslash.


griembauer (approved):
Thanks! Very mergeable from my point of view :)

Issue #180, opened 2021-04-28
Importer: Add support for local file import with file://


neteler:
At time only http requests are supported due to a limitation by design in "python-requests".
It would be important to be able to import geodata files from **local** storage in addition to http requests.

## Problem description
Example: Sentinel-1 data are moved fast to a long term archive, leading to this message:

```
i.sentinel.download ....
Product c0a3c23c-c68f-47d3-9812-8e7b98e19d08 is not online. Triggering retrieval from long term archive.
```

Such a case cannot easily be dealt with in a process chain, hence the separate S-1 scene retrieval and local storage is the only viable solution.

## Needed code changes:
The code here needs to be changed:

https://github.com/mundialis/actinia_core/blob/59cdc55ffab5dea3f8f46f233c3569ffbddb129c/src/actinia_core/resources/common/geodata_download_importer.py#L114

## Potential code solutions

Here some pointers/code snippets:
* https://stackoverflow.com/a/22989322/452464
* https://github.com/dashea/requests-file

anikaweinmann:
Upload of GeoTIFFs in: https://github.com/mundialis/actinia_core/pull/216

griembauer:
Endpoint and tests for new endpoint for raster upload are now implemented. 
TODO: Add raster upload functionality to `importer`-module

PR #259, opened 2021-10-05
Add tests for raster upload endpoint

griembauer:
This PR adds tests for PR #180 , raster upload is tested against reference raster statistics.
Also, a new Assert method is added to `tests/testsuite.py`

developed with @anikaweinmann


mmacata (approved):
Thanks! Did you test the swagger output? I didn't see any problems, just a lot of `"` and `'`.

PR #274, opened 2021-11-12
edited installation steps for actinia without docker

Momen-Mawad:


docs/docs/installation.md
```diff
@@ -117,34 +122,28 @@ and compile it.
  mv LL latlong_wgs84
 ```
 
-4.  Actinia Core must be run within a virtual python3 environment:
+4.  Actinia Core & GRASS GIS 7.8 requires a Python 3 environment for running, so we
+    create one (not needed if only Python 3 is present on the machine).
```
mmacata, line 126: It is often preferred to create a virtual python environment to also keep dependencies independent from other projects and global installation. Not only because of different python version, so please remove this part. (I know that you just moved this part from below but once you are already on it ;) )
```suggestion
    create one.
```

docs/docs/installation.md
```diff
@@ -81,13 +81,17 @@ and compile it.
  git clone https://bitbucket.org/huhabla/temporal-raster-sampling.git t.rast.sample
  cd t.rast.sample
  make
+ ``` 
+> **_HINT:_** run : `sudo apt-get install python-is-python3` If you get the error: `/usr/bin/env: ‘python’: No such file or directory
+` in the logs from `make` command.
```
mmacata, line 86: Please also add to hint that this will change the python version system wide and python2 will not be used anymore. Therefore it is a bit dangerous. How can this be reverted if python2 is needed for old projects? If there is no solution to switch python version with this package (I don't know it), please use https://linuxhint.com/update_alternatives_ubuntu/ instead. It allows to simply configure to which version python should point to.

mmacata (changes requested):
Thanks for the improvement!
I think `python-is-python3` should not be recommended without warning as it changes the behaviour of the `python` command systemwide, I added a dedicated comment.

neteler:
I would just drop all references to Python 2, it is end-of-life for a while (https://www.python.org/doc/sunset-python-2/)

mmacata:
The install instructions were reworked in #287, so I will close this.

PR #275, opened 2021-11-18
STAC in actinia 

joaherrerama:
This pull request implements and integrates STAC in actinia core

### DIFF
--- a/src/actinia_core/core/stac_importer_interface.py
+++ b/src/actinia_core/core/stac_importer_interface.py
@@ -206,7 +206,7 @@ class STACImporter:
                 stac_filter = stac_entry["import_descr"]["filter"]
 
             stac_command = \
-                cls._stac_import(
+                self._stac_import(
                     stac_collection_id=stac_entry_source,
                     semantic_label=stac_semantic_label,
                     interval=stac_interval,
--- a/src/actinia_core/core/stac_importer_interface.py
+++ b/src/actinia_core/core/stac_importer_interface.py
@@ -1,0 +1,168 @@
+# -*- coding: utf-8 -*-
+#######
+# actinia-core - an open source REST API for scalable, distributed, high
+# performance processing of geographical data that uses GRASS GIS for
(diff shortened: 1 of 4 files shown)


src/actinia_core/core/stac_importer_interface.py
```diff
@@ -0,0 +1,214 @@
...
+
+        full_filtered_result = stac_search.json()
+
+        if "features" in full_filtered_result:
+            return full_filtered_result
+        else:
+            raise AsyncProcessError(full_filtered_result)
+
+    @classmethod
+    def get_filtered_bands(self, stac_items, semantic_label):
```
mmacata, line 145: Is a classmethod needed here? Or would it make more sense to use the indication that this method is only used inside this class (`_get_filtered_bands` style). - Same comment goes to all methods here except `get_stac_import_download_commands`. And to my knowledge it is a convention that `cls` is used instead of `self` when adressing this method, but I am no expert on this.

Maybe you can indicate why you used a classmethod here and we can discuss what makes more sense!?

mmacata (changes requested):
Thank you for this great feature!
Please update year and typo and let's discuss about the classmethods. (No real preference here, just that it is not used much in actinia_core yet besides the tests).

And please also update [API docs](https://github.com/mundialis/actinia_core/blob/main/src/actinia_core/models/process_chain.py#L122) in enumeration and possible new parameters and descriptions.

src/actinia_core/core/stac_importer_interface.py
joaherrerama: I couldn't handle `@classmethod ` and `self` functions to work together. So far what I understood is that you can not call a self function inside a `@classmethod `, one should either `@staticmethods ` and  `self` functions  for our purpose and the class method should be focus on object characteristics. (somehow it makes sense to me)

mmacata: Ok, makes sense.

mmacata (approved):
Thanks, please merge!

PR #276, opened 2021-11-22
Test for STAC implementation

joaherrerama:


tests/test_async_processing_import_export.py
```diff
@@ -323,6 +404,46 @@ def test_sentinel_import_error(self):
         self.waitAsyncStatusAssertHTTP(rv, headers=self.admin_auth_header,
                                        http_status=400, status="error")
 
+    # Test for STAC
+    @unittest.skipIf('GOOGLE_APPLICATION_CREDENTIALS' not in os.environ and 'GOOGLE_CLOUD_PROJECT' not in os.environ,
+                     "Test is skipped because 'GOOGLE_APPLICATION_CREDENTIALS' and 'GOOGLE_CLOUD_PROJECT' not set")
```
mmacata, line 409: `GOOGLE_APPLICATION_CREDENTIALS` and `GOOGLE_CLOUD_PROJECT` is not needed here. Instead, could you test if the stac plugin is installed and skip the tests if not? (Also valid for below tests).

tests/test_async_processing_import_export.py
```diff
@@ -323,6 +404,46 @@ def test_sentinel_import_error(self):
...
+    def test_stac_import(self):
+
```
mmacata, line 411: Could you please add a sentence, saying what this test is doing? We agreed upon starting this for all new tests, so no great example until now, but then we have a good example ;)

mmacata (commented):
Thank you for this update. I only have small comments.

tests/test_async_processing_import_export.py
```diff
@@ -32,9 +32,15 @@
 except ModuleNotFoundError:
     from test_resource_base import ActiniaResourceTestCaseBase, URL_PREFIX
 
+try:
+    import actinia_stac_plugin
+    no_stac_plugin = False
+except Exception:
+    no_stac_plugin = True
```
joaherrerama, line 39: This was the best option that cam up to my mind on how to verify the installation. Any suggestion to make it cleaner?

mmacata: Fine for me. What do you think @anikaweinmann ?

anikaweinmann: For me it is also fine

anikaweinmann (approved):

mmacata (approved):

PR #278, opened 2021-11-23
Use separate repo for api docs

mmacata:
To be able to release the actinia API individually from the code, this PR starts to move openAPI descriptions to a separate repository https://github.com/mundialis/actinia-api.
This PR

- moves some of the API docs to actinia-api
- replaces their import
- adjusts the build to make use of the separate repository
- adjusts the build for updated plugins regarding `api/v1` -> `api/v2` changes
- add response to all old `api/v1` endpoints. Example:
```
$ curl http://127.0.0.1:8088/api/v1/locations/nc_spm_08/mapsets | jq
[
  {
    "links": [
      "http://127.0.0.1:8088/api/v2/locations/nc_spm_08/mapsets"
    ],
    "message": "Are you looking for the current api version? Change 'v1' to 'v2' in the URL.",
    "status": "Not found"
  },
  404
]

```

Follow up:
-  #279


src/actinia_core/version.py
```diff
@@ -133,3 +135,20 @@ def version():
...
+    version is installed, this endpoint will be overwritten by a proxy.
+
+    Returns: Response
+
+    """
+    url = request.url_root.strip('/') + URL_PREFIX + '/' + actinia_path
+
+    return make_response(jsonify(LinkResponseModel(
+            status="Not found",
+            message="Are you looking for the current api version?",
```
griembauer, line 153: Maybe it helps to specify in the message which part of the URL can be adapted. Since so far it had always been `.../api/v1/...`, it might not occur to everyone that it is the `v1` that needs to be changed (Although quite obvious, but you never know ;))

griembauer (approved):
Thank you, just one minor comment.

src/actinia_core/version.py
juleshaas: yes, from a beginner's point of view I would definitely agree, that I probably wouldn't get it right away...

mmacata: Ok, I changed it in the description above accordingly.

src/actinia_core/version.py
```diff
@@ -133,3 +135,20 @@ def version():
     info['running_since'] = find_running_since_info()
 
     return make_response(jsonify(info), 200)
+
+
+# Return a hint that this version is outdated
+@flask_app.route("/api/v1/<path:actinia_path>")
```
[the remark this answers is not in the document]
mmacata: It could but usually a proxy is used to direct requests to the same root URL to the corresponding app. If actinia v1 is also installed, the proxy needs to take care about corresponding routing and to overwrite these endpoints.

Do you mean OLD_VERSION as variable in the code? I would argue against it that if we keep it as is, we can still decide in a later change if we want to add an additional endpoint for the newer older version or just return the hint for the latest deprecated version.
If you mean to replace `v1` in the URL with `<old_version>`, I think it is too risky because it provides a duplicate endpoint for all endpoints because it also fits to `v2`. In my current test it did work but I don't know if flask behaves predictable in this case.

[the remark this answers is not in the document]
anikaweinmann: I mean OLD_VERSION as variable. I do not think that we need vor each old verson the hint, but it is also fine for me to keep it like it is

anikaweinmann (approved):

PR #281, opened 2021-11-26
More secure path mangling

marcjansen:
This is a follow up from the excellent #280. I think the security might even be improved by the following PR.

I'd really like a review from @anikaweinmann or @mmacata on this; this is open for discussion.

### DIFF
--- a/src/actinia_core/core/utils.py
+++ b/src/actinia_core/core/utils.py
@@ -28,6 +28,7 @@ Utils file with common util functions
 import os
 from actinia_core.core.common.process_object import Process
 from actinia_core.core.common.exceptions import SecurityError
+from actinia_core.core.common.config import global_config
 
 __license__ = "GPLv3"
 __author__ = "Sören Gebbert, Anika Weinmann"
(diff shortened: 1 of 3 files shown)


marcjansen:
Hmm. Probably some occasions missed. Hopefully just that.

src/actinia_core/core/utils.py
```diff
@@ -49,11 +50,22 @@ def os_path_normpath(path_parts):
...
         raise SecurityError(f"Used path '{path}' is insecure")
     return path
 
 
+def path_is_in_allowed_paths(path):
+    # TODO we should double check whether they are all needed, better would be one
+    # single location
+    allowed_starts = [
+        global_config.GRASS_USER_DATABASE,
+        global_config.GRASS_DATABASE,
```
anikaweinmann, line 63: Carmen is right that `GRASS_DATABASE` is only okay if read and not if you want to change something in it. Maybe you should still distinguish that somehow?

src/actinia_core/core/utils.py
```diff
@@ -49,11 +50,22 @@ def os_path_normpath(path_parts):
...
+        global_config.GRASS_RESOURCE_DIR
+    ]
+    return any(path.startswith(allowed) for allowed in allowed_starts)
```
anikaweinmann, line 66: Theoretically the `startswith` could be unsafe here again, if  `path="GRASS_RESOURCE_DIR/../../something_else`.

src/actinia_core/core/utils.py
marcjansen: Well it is used in [these lines of PR 280](https://github.com/mundialis/actinia_core/pull/280/files#diff-ee1d2bbb5ac0ad6fc98c8c9b1f98cb33232ee8d70320349b17362de44c629edcR217), that is why I included it.

src/actinia_core/core/utils.py
anikaweinmann: `os.path.normpath("GRASS_RESOURCE_DIR/../../something_else")` leads to `../something_else`, so the normpath is needed that this function `path_is_in_allowed_paths` is working. Maybe you can mention that path should be a path which was corrected with `os.path.normpath`.

src/actinia_core/core/utils.py
anikaweinmann: I am not sure if it makes sense to add some `r`/`w` parameter to the functions and split the `allowed_starts`.

marcjansen: Ah, I now understand what you mean. I can update this as suggested.

src/actinia_core/core/utils.py
marcjansen: But if `os.path.normpath("GRASS_RESOURCE_DIR/../../something_else")` leads to `../something_else` the function would return false, as it does not start with any of the allowed prefixes. I can of course add a note

anikaweinmann:
@marcjansen do you want to add a `r`/`w` parameter to the functions? Otherwise it is also fine for me to merge it as it is.

anikaweinmann (approved):

marcjansen:
Sorry for not giving this more attention. I think the general idea is still valid, but I am not sure whether this catches all relevant cases. 

I see this has an approving review, and I can try to rebase this, if you still want it. I can also live with you guys and girls closing this PR with unmerged commits. Please go ahead as you see fit.

anikaweinmann:
@marcjansen you are welcome to merge and rebase

src/actinia_core/core/utils.py
marcjansen: I have updated the PR; `ensure_valid_path` now accepts an optional second parameter, the intent ("r", "w" or "rw", Default is "r")

marcjansen:
> @marcjansen you are welcome to merge and rebase

I have done that and also added support for intents (`r`, `w` or `rw`) when `ensure_valid_path` is called.

Please have another close look at what is now in this PR. Where the method is called, I have added an intent where I was more or less sure, but these shpould please be double checked by someone with more know-how of the actual intent.

marcjansen:
anybody @mundialis-dev ?

marcjansen:
Friendly ping @mmacata @anikaweinmann

anikaweinmann:
Thank you, for me it look good. Please merge.

marcjansen:
Thanks for the review.

PR #284, opened 2021-12-03
Enable debugging via vscode

mmacata:
This PR adds `.vscode` files for more elegant local debugging. The current local dev setup workflow, described in the [documentation](https://github.com/mundialis/actinia_core/blob/main/docker/README.md#local-dev-setup-with-docker) is still valid and working, although Dockerfile + compose file were adjusted.

New debugging workflow with [vscode](https://code.visualstudio.com/) is as follows (also explained in the README):
- open up `code` and make sure the workspace is set to `actinia_core`
- press `F5` and wait a few seconds - a browser window should be opened pointing to the version endpoint
- For debugging tips, [read the docs](https://code.visualstudio.com/Docs/editor/debugging#_debug-actions).

Some debugging features:
- Set breakpoints via click in code editor
- After local changes, the debugger reloads automatically and restarts actinia
- On initial debug startup, a new browser window pointing to `http://localhost:8088/api/v2/version` opens automatically. 

Some things to enhance (in follow up PR):
- make live reload after local changes a bit faster, currently it takes ~10 seconds, which is still better as manual reload
- include debugging options for actinia-plugins - currently, still the "old" debugging workflow needs to be used.


docker/docker-compose-dev.yml
```diff
@@ -3,27 +3,29 @@ services:
 
   actinia:
     build:
-      context: actinia-core-dev/
+      context: ..
```
griembauer, line 6: Just for my understanding: This means we look at the entire `actinia_core` directory?

docker/docker-compose-dev.yml
```diff
@@ -41,21 +43,8 @@ services:
...
-  #   image:  elasticsearch:7.4.0
-  #   environment:
-  #   - discovery.type=single-node
-  #
-  # kibana:
-  #   image: kibana:7.4.0
-  #   ports:
-  #     - "5601:5601"
+networks:
+  actinia-dev:
```
griembauer, line 50: Just for my understanding: Here we define the network such that both `actinia` and `redis` can be in the same network? But as it is here, no further configuration of the network is required?

griembauer (approved):
Great addition! I can't really assess the code but I do like the new feature very much! ;)

docker/actinia-core-dev/Dockerfile
```diff
@@ -1,9 +1,22 @@
-FROM mundialis/actinia-core:g78-latest-alpine
+FROM mundialis/actinia-core:g79-latest-alpine
```
neteler, line 1: Is there meanwhile also a g80 image available? g79 has been moved ...

docker/docker-compose-dev.yml
mmacata: Yes, the whole repository is used as build context now. This is needed because the Dockerfile was adjusted to copy the whole source code inside and it doesn't work exclusively by mounting the code anymore.

docker/docker-compose-dev.yml
mmacata: Correct, when nothing is specified, all services inside the compose file are using the same network on startup.
Here a docker network is specified explicitly. When redis is started, it will automatically be created and used without further configuration. When actinia is started via debugger, this network is also specified: https://github.com/mundialis/actinia_core/blob/909d0cd41e14862fbea72e14504c2200c64b674a/.vscode/tasks.json#L44 
Docker compose names the network like `<name-of-folder-where-compose-file-is-inside>_<name-of-network>` ).

PR #285, opened 2021-12-03
New splitup in unit and integration tests

anikaweinmann:
The tests are now splitted into unittests marked with `@pytest.mark.unittest` and others (= integration tests).
As unittest example there are added few tests for the version but any other test method can be marked as unittest.
For development of the tests you can add a marker `@pytest.mark.dev` to specific tests and run only the marked tests with `make devtest` see [tests/README.md](https://github.com/mundialis/actinia_core/blob/tests/tests/README.md).

The integration and unittests are both added to the github actions.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -16,6 +16,8 @@ docker build -f docker/actinia-core-tests/Dockerfile -t actinia-test .
 ```
 3. To run only a few tests you can mark the tests for development with
 `@pytest.mark.dev` and add `import pytest` to the file.
+(For best practice examples on the use of pytest-decorators, see `unittests/test_version.py`)
+
 
 4. Start the docker container and mount your `tests` folder:
 
@@ -34,7 +36,7 @@ docker run -it actinia-test:latest -i
 make test
 
 # execute only integration tests (not unittests)
-make notunittest
+make integrationtest
 
 # execute only unit tests
 make unittest
@@ -45,8 +47,6 @@ make devtest
 # If you added a debugger to your test it will stop there.
 # After making changes to the test, you need to close and restart the docker container (docker run ...) before testing again.
 ```
-
-When you are done, add your new test(s) to `actinia_core/tests`. You can run the entire testsuite including your new tests by rebuilding the docker image and running the container without mounting (see 4.)
 
 
 ## Problems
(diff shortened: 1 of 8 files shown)


tests/README.md
```diff
@@ -14,29 +14,34 @@ When writing new tests it is useful to run selected tests locally and isolated.
 ```
 docker build -f docker/actinia-core-tests/Dockerfile -t actinia-test .
 ```
-3. Create a local `tmp_tests` folder and fill it with the files:
+3. To run only a few tests you can mark the tests for development with
+`@pytest.mark.dev` and add `import pytest` to the file.
```
griembauer, line 18:
```suggestion
`@pytest.mark.dev` and add `import pytest` to the `.py` file/s with the tests you want to run.
```
--> if I understand correctly?

tests/README.md
```diff
@@ -14,29 +14,34 @@ When writing new tests it is useful to run selected tests locally and isolated.
...
 ```
 
 5. To execute the test(s) run:
 
 ```
+# execute all tests
 make test
 
+# execute only integration tests (not unittests)
+make notunittest
```
griembauer, line 37: just for my understanding: Shouldn't this be 
`make integrationtest`, because here the `Makefile` is called and not directly `tests_with_redis.sh`?

tests/README.md
```diff
@@ -14,29 +14,34 @@ When writing new tests it is useful to run selected tests locally and isolated.
...
+
+# execute only unit tests
+make unittest
+
+# execute tests marked with '@pytest.mark.dev'
+make devtest
+
```
griembauer, line 44: Not sure where to put it, but somewhere in the test-README I think the new ` tests/unittests/test_version.py` should be mentioned as an example for pytest-decorator usage, because I find this `pytest.mark.parametrize`-thing very helpful.  Something like:
```
For best practice examples on the use of pytest-decorators, see `unittests/test_version.py`
```

griembauer (approved):
Thanks a lot, very helpful new insights into `unittests`! I added just some minor questions and suggestions.

tests/README.md
anikaweinmann: I am not sure if this are really integration tests or not? But I can rename it, if we decide that this are all integration tests

marcjansen:
Plus `1` for the general direction. Not a proper review, but really happy about this.

mmacata (approved):

PR #289, opened 2021-12-16
update actinia-api version

mmacata:
See https://github.com/mundialis/actinia-api/releases


mmacata:
As only tiny change, I will merge directly.

PR #295, opened 2022-01-06
adapt s2 importer to work without Google BigQuery

griembauer:
This PR adapts the actinia `importer` process for the importing of Sentinel-2 data. Sentinel-2 data (indicated by a scene-ID) can now be imported directly from Google Cloud Storage without providing any credentials using `i.sentinel.download` and `i.sentinel.import` in the background. This PR currently uses the standard implementation of the `importer`: each `import_description` can be used to import exactly ONE raster map. For Sentinel-2 data, this means that each band has to be specified individually. 

Note: The current implementation (query via Google Cloud BigQuery interface etc.) is not removed in this PR, the new functionality without login is simply added.

perspective TODOs:

- implement extent (bbox): import processes are automatically put before grass (or other) processes in a process chain. This means that no region can be set via a grass module before running the import. Limiting the import data to a certain extent has to be done via the [extent parameter](https://github.com/mundialis/actinia_core/blob/main/src/actinia_core/models/process_chain.py#L167) of the `importer`. This still needs to be implemented. 
- correct null value: When importing a scene, `i.sentinel.import` reprojects the map if it does not match the current projection. This may interfere with 0-values, that should be treated as NULL-values and removed via `r.null` or `i.zero2null`. This problem should probably be solved in `r.import`
- improve `i.sentinel.download` such that it can directly download individual bands (and required metadata) instead of the entire scene, (and maybe even cut it to the current region)
- implement query such that all S2 data from region and time of interest are imported (e.g. using `i.sentinel.coverage` + `t.sentinel.import`). This is currently implemented with the BigQuery approach and should be replaced.

Example PC:
```
{
 "list": [
          {"id": "importer_1",
          "module": "importer",
          "inputs": [{"import_descr": {"source": "S2B_MSIL1C_20210809T101559_N0301_R065_T32UNB_20210809T124430",
                                       "type": "sentinel2",
                                       "sentinel_band": "B04"},
                      "param": "map",
                      "value": "B04"},
                     {"import_descr": {"source": "S2B_MSIL1C_20210809T101559_N0301_R065_T32UNB_20210809T124430",
                                       "type": "sentinel2",
                                       "sentinel_band": "B08"},
                      "param": "map",
                      "value": "B08"}]},
          {
      			"id": "dummy_region",
      			"comment": "Set the region to raster",
      			"module": "g.region",
      			"inputs": [
      				{
      					"param": "raster",
      					"value": "B04"
      				}
      			],
      			"flags": "p"
      		},
          {
      			"id": "dummy_mapcalc",
      			"comment": "just a dummy module",
      			"module": "r.mapcalc",
      			"inputs": [
      				{
      					"param": "expression",
      					"value": "test_raster = B04 + B08"
      				}
      			]
      		}
         ],
 "version": "1"
}
```


tests/test_sentinel2_processing_commands.py
```diff
@@ -95,6 +96,45 @@ def test_download_import_commands(self):
 
         # self.assertTrue(len(result) == 2)
 
+    def test_download_import_commands_noquery(self):
+        gsqi = Sentinel2Processing(product_id="S2A_MSIL1C_20170212T104141_N0204_R008_T31TGJ_20170212T104138",
+                                   bands=["B12", "B08"], download_cache="/tmp",
```
neteler, line 101: Is the currently hardcoded `download_cache="/tmp"` a good idea?
I'd prefer it to be set via config file or ENV variable (whatever is better).

tests/test_sentinel2_processing_commands.py
```diff
@@ -95,6 +96,45 @@ def test_download_import_commands(self):
...
+        result, maps = gsqi.get_sentinel2_download_process_list_without_query()
+        # there should be just one command (i.sentinel.download)
+        self.assertTrue(len(result) == 1, ("Download process list consists of "
+                                           "more than one processes"))
+        self.assertEqual(result[0].executable, "i.sentinel.download",
+                         ("Download process executable is not "
+                          "i.sentinel.download"))
+        ref_params = ['datasource=GCS',
+                      ('query=identifier=S2A_MSIL1C_20170212T104141'
+                       '_N0204_R008_T31TGJ_20170212T104138'), 'output=/tmp']
```
neteler, line 114: `output=/tmp`  - see comment above

tests/test_sentinel2_processing_commands.py
griembauer: Note that this test does not actually download anything. It is merely a test whether the `importer`- module is interpreted correctly and leads to the correct download/import commands. However, it is indeed nicer to make the temporary directory configurable. Since no download actually happens, it is now a class attribute.

src/actinia_core/core/common/process_chain.py
```diff
@@ -279,6 +278,57 @@ def _get_sentinel_import_command(self, entry):
 
         return sentinel_commands
 
+    def _get_sentinel_import_commands(self, entries):
```
anikaweinmann, line 281: Maybe you could split this rather large method again? Or pull out the first part as `_get_setinel_scenen_info` or similar, so that it is not so big.

src/actinia_core/core/common/sentinel_processing_library.py
```diff
@@ -32,6 +32,7 @@
 from .exceptions import AsyncProcessError
 from .process_object import Process
 
+
 __license__ = "GPLv3"
 __author__ = "Sören Gebbert"
 __copyright__ = "Copyright 2016-2018, Sören Gebbert and mundialis GmbH & Co. KG"
```
anikaweinmann, line 38:
```suggestion
__copyright__ = "Copyright 2016-2022, Sören Gebbert and mundialis GmbH & Co. KG"
```

src/actinia_core/core/common/sentinel_processing_library.py
```diff
@@ -32,6 +32,7 @@
 from .exceptions import AsyncProcessError
 from .process_object import Process
 
+
 __license__ = "GPLv3"
 __author__ = "Sören Gebbert"
```
anikaweinmann, line 37:
```suggestion
__author__ = "Sören Gebbert, Guido Riembauer"
```

tests/test_sentinel2_processing_commands.py
```diff
@@ -28,6 +28,7 @@
 from actinia_core.core.common.config import global_config
 import os
 import unittest
+import pytest
 
 __license__ = "GPLv3"
 __author__ = "Sören Gebbert"
```
anikaweinmann, line 34:
```suggestion
__author__ = "Sören Gebbert, Guido Riembauer"
```

tests/test_sentinel2_processing_commands.py
```diff
@@ -95,6 +96,45 @@ def test_download_import_commands(self):
 
         # self.assertTrue(len(result) == 2)
 
+    def test_download_import_commands_noquery(self):
```
anikaweinmann, line 103: Isn't this a unit test?

tests/test_sentinel2_processing_commands.py
```diff
@@ -28,6 +28,7 @@
 from actinia_core.core.common.config import global_config
 import os
 import unittest
+import pytest
```
anikaweinmann, line 31: `pytest` is not used?

src/actinia_core/core/common/process_chain.py
griembauer: This method is now split up

tests/test_sentinel2_processing_commands.py
griembauer: Thanks, now it is due to the unittest decorator.

tests/test_sentinel2_processing_commands.py
griembauer: Yes, you are right. The `pytest` decorator has been added and it is moved to the `unittest` dir

anikaweinmann (approved):
Thank you.

Issue #244, opened 2019-11-21
importer: add parameters to support raster resampling method and target resolution


neteler:
At time the importer runs `r.import` always with "nearest" (neighbor) resampling which is a problem when importing surfaces (temp, elev, ...) in case that internally a reprojection takes place.

Proposed solution:

- add two parameters to the actinia "importer" interface to support
    - **method**: nearest (default), bilinear_f (so, maybe not all offered by [r.import](https://grass.osgeo.org/grass78/manuals/r.import.html))
    - **resolution**: numeric value

which internally constructs this code:
- `resample=bilinear_f resolution=value resolution_value=XX`

to be added here:

https://github.com/mundialis/actinia_core/blob/eb8bca301010a2ca71a40def4c0a962e26eb7fd6/src/actinia_core/resources/common/geodata_download_importer.py#L236

Note: Two parameters are sufficient since:
- left empy: "nearest" is used as before
- only **resample** is set: propagate "resample" to r.import
- only **value** is set: use "nearest
- both **resample** and **value** are set: propagate "resample" and "resolution_value" to r.import

PR #298, opened 2022-01-07
Importer resampling and resolution

juleshaas:
In this PR the parameters `resample`, `resolution` and `resolution_value` are added to the raster importer.

Resampling can be set to the options available in GRASS r.import:

    nearest: nearest neighbor (default)
    bilinear: bilinear interpolation
    bicubic: bicubic interpolation
    lanczos: lanczos filter
    bilinear_f: bilinear interpolation with fallback
    bicubic_f: bicubic interpolation with fallback
    lanczos_f: lanczos filter with fallback

Resolution can be set to:

    estimated: estimated resolution (default)
    value: user-specified resolution
    region: current region resolution

When `resolution` is set to `value`, the parameter `resolution_value` must also be defined.

Example process chain (part):
```
{
      "id": "importer_1",
      "module": "importer",
      "inputs": [
        {
          "import_descr": {
            "source": "https://url-to-rasterfile.tiff",
            "type": "raster",
            "resample": "bilinear_f",
            "resolution": "value",
            "resolution_value": "100"
          },
          "param": "map",
          "value": "name_of_map"
        }
      ]
    },
```

This PR solves issue #244.


src/actinia_core/core/geodata_download_importer.py
```diff
@@ -215,22 +215,35 @@ def get_file_rename_command(file_path, file_name):
...
     @staticmethod
-    def get_raster_import_command(file_path, raster_name):
+    def get_raster_import_command(file_path, raster_name, resample=None,
+                                  resolution=None, resolution_value=None):
         """Generate raster import process list that makes use of r.import
 
         Args:
             file_path:
             raster_name:
+            resample:
```
anikaweinmann, line 225: Maybe it wouldn't hurt here to add the datatype and maybe a descriptive sentence like https://github.com/mundialis/actinia_core/blob/main/src/actinia_core/core/common/api_logger.py#L134-L136

anikaweinmann (approved):

src/actinia_core/core/geodata_download_importer.py
juleshaas: I added some description here , however not sure about the datatype for `resolution_value`. The value is given as string but needs to be convertable to float to use in r.import.

anikaweinmann: For the method is has to be a string. Otherwise you has to change `executable_params.append("resolution_value=%s" % resolution_value)` to ` executable_params.append("resolution_value=%d" % resolution_value)`.

src/actinia_core/core/geodata_download_importer.py
```diff
@@ -215,22 +215,39 @@ def get_file_rename_command(file_path, file_name):
...
     @staticmethod
-    def get_raster_import_command(file_path, raster_name):
+    def get_raster_import_command(file_path, raster_name, resample=None,
+                                  resolution=None, resolution_value=None):
         """Generate raster import process list that makes use of r.import
 
         Args:
-            file_path:
-            raster_name:
+            file_path (str): The URL of the raster file to import
```
anikaweinmann, line 223:
```suggestion
            file_path (str): The path of the raster file to import
```

anikaweinmann (approved):

src/actinia_core/core/geodata_download_importer.py
juleshaas: Thanks, I set it to `str` then.

src/actinia_core/core/geodata_download_importer.py
```diff
@@ -215,22 +215,39 @@ def get_file_rename_command(file_path, file_name):
...
         Returns:
             Process
 
         """
+        executable_params = ["input=%s" % file_path,
+                             "output=%s" % raster_name,
+                             "--q"]
+
+        if resample is not None:
+            executable_params.append("resample=%s" % resample)
```
griembauer, line 242: Would there be any benefit in checking if `resample` has one of the allowed values `"nearest", "bilinear"`, etc.? This is not the case right now, and I reckon then `r.import` fails with a proper error message. It might still be worth it, because then the process fails directly and not when it comes to the importing step (although all importing steps are performed at the beginning of the PC)

griembauer (commented):
Just one small comment, otherwise please merge. You could also add a test for the new functionality (but not mandatory of course) ;)

neteler:
> You could also add a test for the new functionality

Yes, it would be good to add a test. Material may be used from here:
https://github.com/OSGeo/grass/tree/main/scripts/r.import/testsuite/

mmacata:
Thanks! Please also add new parameters to API docs: https://github.com/mundialis/actinia_core/blob/main/src/actinia_core/models/process_chain.py#L86

src/actinia_core/core/geodata_download_importer.py
juleshaas: checks have been added in commit [66de85f3bf17dbe048b22336fa07ec6037f5653f](https://github.com/mundialis/actinia_core/pull/298/commits/66de85f3bf17dbe048b22336fa07ec6037f5653f)

juleshaas:
> Thanks! Please also add new parameters to API docs: https://github.com/mundialis/actinia_core/blob/main/src/actinia_core/models/process_chain.py#L86

added to [src/actinia_core/models/process_chain.py](https://github.com/mundialis/actinia_core/pull/298/files#diff-d01f581e02e0a0c986929c458043248cffb52a2383656abacdedb68828db9933)

juleshaas:
> > You could also add a test for the new functionality
> 
> Yes, it would be good to add a test. Material may be used from here: https://github.com/OSGeo/grass/tree/main/scripts/r.import/testsuite/

added a follow-up issue for tests: #328

anikaweinmann (approved):

metzm:
Can you add somewhere a note that the `resolution_value` must be in units of the target coordinate reference system, not in map units?

PR #299, opened 2022-01-13
Retry webhook

anikaweinmann:
If a `finished_webhook` is set this is informed if the process is finished or has an error. Unit now this `finished_webhook` was only called once and when the connection fails the `finished_webhook` never get to know that the process is not running anymore.
This change make it configurable, that the `finished_webhook` can be retried. For this you can set in the `actinia.cfg` a WEBHOOK section like this:
```
[WEBHOOK]
WEBHOOK_RETRIES = 6
WEBHOOK_SLEEP = 10
```


mmacata (approved):
I like the new script named `webhook-server-broken` :upside_down: but makes totally sense for the test.
Thank you, please merge!

PR #308, opened 2022-01-28
allow '&' in parameters for more modules

metzm:
all modules where a parameter can be a mapcalc expression must allow '&' in the argument string


anikaweinmann (approved):
Thank you. Please merge

PR #313, opened 2022-02-23
Add pipeline to add python wheel to release assets

mmacata:
Title says it all.


mmacata:
As same in actinia-stac-plugin, I will merge directly.

PR #318, opened 2022-03-15
STAC Importer Issues

joaherrerama:
- [x] Adding `t.create` to create the STDR
- [x] Replacing `r.in.gdal` to `r.import `to reproject automatically the images coming from the STAC Item 
- [x] Registering the image into the STDR - Associating the image date
- [x] Registering the semantic label coming from the STAC ITEM
- [x] Setting value providing by the user
- [x] Supporting POST and GET HTTP method in /searchsi
- [x] Setting`/vsis3` for S3 bucker GDAL reading and `/vsicurl `for other formats


src/actinia_core/core/stac_importer_interface.py
```diff
@@ -24,7 +24,6 @@
 """
 This code shows the STACImporter model which help actinia core
```
neteler, line 25:
```suggestion
This code shows the STACImporter model which helps actinia core
```

src/actinia_core/core/stac_importer_interface.py
```diff
@@ -142,26 +144,69 @@ def _stac_import(self, stac_collection_id=None, semantic_label=None,
...
-                for name_id, url in value.items():
+                for name_id, url, item_date in value.items():
 
                     output_name = stac_name + "_" + key + "_" + name_id
 
                     # From Here Onwards, the Process build starts
                     exec_params = ["input=%s" % "/vsicurl/"+url,
                                    "output=%s" % output_name,
-                                   "-o"]
+                                   "extent=region"]
```
neteler, line 183: Should also `resolution=region` be used? At time, an estimated resolution value is used.

joaherrerama: Yes, I didn't know which were more convenient in this case. But it makes sense to me, to set resolution as regional 👍🏽

metzm: I tend to use the estimated resolution because e.g. with Sentinel-2 the different bands have different native resolutions. Resampling all bands to the same resolution might not be desired.

src/actinia_core/core/stac_importer_interface.py
```diff
@@ -142,26 +144,70 @@ def _stac_import(self, stac_collection_id=None, semantic_label=None,
...
 
                     p = Process(
                         exec_type="grass",
-                        executable="r.in.gdal",
+                        executable="r.import",
                         executable_params=exec_params,
-                        id=f"r_gdal_{os.path.basename(output_name)}",
+                        id=f"r_import_{os.path.basename(output_name)}",
                         skip_permission_check=True
                     )
```
metzm, line 193: After import, you can set the semantic label with `r.support`. `t.register` will then automatically use this semantic label and insert the raster map with metadata accordingly in the TGIS db.

src/actinia_core/core/stac_importer_interface.py
joaherrerama: This resolution refers to Spatial resolution, right? If true, resampling the resolution might be convenient in terms of operational processing inside the Location. However, it can cause a loss of information if the image is upscaling.

src/actinia_core/core/stac_importer_interface.py
joaherrerama: is it [`r.support`](https://grass.osgeo.org/grass78/manuals/r.support.html)? what's the difference with [`r.semantic.label`](https://grass.osgeo.org/grass80/manuals/r.semantic.label.html)

neteler: It needs to be v8x: https://grass.osgeo.org/grass80/manuals/r.support.html#set-semantic-label 
Which of the two is preferred, I cannot say.

src/actinia_core/core/stac_importer_interface.py
```diff
@@ -95,37 +93,36 @@ def _apply_filter(stac_root_search, stac_name, interval, bbox, filter):
 
         full_filtered_result = stac_search.json()
 
-        if "features" in full_filtered_result:
+        if "features" in full_filtered_result\
```
metzm, line 96: whitespace before `\`, check with black or flake8
```suggestion
        if "features" in full_filtered_result \
```

src/actinia_core/core/stac_importer_interface.py
```diff
@@ -95,37 +93,36 @@ def _apply_filter(stac_root_search, stac_name, interval, bbox, filter):
...
-                        if value["eo:bands"][0]["common_name"] in semantic_label:
-                            band_name = value["eo:bands"][0]["name"]
-                            if band_name not in band_roots:
-                                band_roots[band_name] = {}
-                            feature_id = feature["id"]
-                            item_link = feature["assets"][band_name]["href"]
-                            band_roots[band_name][feature_id] = item_link
-                        elif value["eo:bands"][0]["name"] in semantic_label:
+                        if value["eo:bands"][0]["common_name"] in semantic_label\
+                                or value["eo:bands"][0]["name"] in semantic_label\
```
metzm, line 112: whitespace before `\`, check with black or flake8
```suggestion
                        if value["eo:bands"][0]["common_name"] in semantic_label \
                                or value["eo:bands"][0]["name"] in semantic_label \
```

src/actinia_core/core/stac_importer_interface.py
```diff
@@ -95,37 +93,36 @@ def _apply_filter(stac_root_search, stac_name, interval, bbox, filter):
...
                             item_link = feature["assets"][band_name]["href"]
-                            band_roots[band_name][feature_id] = item_link
+                            band_roots[band_name]["name_id"] = feature_id
+                            band_roots[band_name]["url"] = item_link
+                            band_roots[band_name]["datetime"] = item_date
         return band_roots
 
-    def _stac_import(self, stac_collection_id=None, semantic_label=None,
-                     interval=None, bbox=None, filter=None):
+    def _stac_import(self, stac_collection_id=None, semantic_label=[],
```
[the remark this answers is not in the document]
joaherrerama: I did it as the empty list is the default input. And a user can set no `semantic label` and the empty list to allow mapping all the bands of the item. Maybe should the conditional consider `None` and `[ ]` as entries for all bands?

requirements.txt
```diff
@@ -52,3 +51,5 @@ geopandas
 
 # TODO: remove threadpoolctl and joblib if not needed anymore for scikit-learn
 # google-cloud-bigquery needs libffi-dev
+
+# scikit-learn
```
mmacata: As this change is not relevant for this PR, please undo, as it changes the history (and especially in these requirement files, a git blame is often very helpful).

src/actinia_core/core/stac_importer_interface.py
```diff
@@ -164,11 +175,15 @@ def _stac_import(self, stac_collection_id=None, semantic_label=[],
                 item_id = value["name_id"]
                 output_name = f"{strd_name}_{item_id}_{key}"
 
+                url_prefix = "/vsicurl/"
+                # Checking if the URL belongs to S3
+                if "s3:" in value["url"]:
+                    url_prefix = "/vsis3/"
```
mmacata, line 181: That's all and it works? Cool!

mmacata (approved):
Thanks :)
Only one wish that the "no change" in the requirements is removed, then pls merge directly.

Issue #191, opened 2021-05-20
Split code into job receiver + job executor


mmacata:
... to prepare for additional job manager later.

Issue #189, opened 2021-05-20
Split up large functions and modules


mmacata:
title says it all

PR #322, opened 2022-03-18
Refactor rest - part 2

mmacata:
See also MR #320 and issues #190, #189 and #191

# General refactoring thoughts:
This PR splits up the `rest` modules into:
- the class used with flask which stays in the existing `rest` module
For `raster_colors` e.g. `class SyncPersistentRasterColorsResource(ResourceBase)`
- the classes used for processing, separated into `persistent` and `ephemeral` processing as well as `ephemeral_with_export` and `ephemeral_renderer_base`. They are moved to `processing.actinia_processing`. Perspectively this will become a separate repository.
For `raster_colors` e.g. `class EphemeralRasterColorsOutput(EphemeralProcessing)` and `class PersistentRasterColorsRules(PersistentProcessing)`
- the methods in between which are used to write processing jobs into the queue and then to read from it are moved to `processing.common`
For `raster_colors` e.g. `def start_job_colors_out(*args)` and `def start_job_from_rules(*args)`

# Effects on the queue (local / redis)
Regarding the queue - the methods in between mentioned above can be used in 3 different ways:
- in the job receiver with `QUEUE_TYPE = local`. Here, actinia_processing is needed. If the import fails, actinia won't start up with `ModuleNotFoundError: No module named 'actinia_core.processing' - No actinia_processing found but required for local queue! [in /usr/lib/python3.8/site-packages/werkzeug/_internal.py:113]`
- in the job receiver with `QUEUE_TYPE = redis` - import doesn't matter as no processing done here
- in the job executor. actinia_processing is the essential part and import is highly relevant.

# One note about sorting inside `actinia_processing`
The modules are sorted by inheritance into the subfolders
- `ephemeral`,
- `ephemeral_with_export`,
- `persistent` and 
- `ephemeral_renderer_base`. 

This means eg. that the module `persistent_processing.py` is inside the folder `ephemeral` because it inherits from `EphemeralProcessing` directly: `class PersistentProcessing(EphemeralProcessing)`. The module `ephemeral_processing`, from which all other classes inherit directly or indirectly, is on top level next to the folders.

# WARNING
Last thing: again a WARNING that some actinia plugins are affected by move of
- ephemeral_processing_with_export
- ephemeral_processing
- persistent_processing

Alle of them are adjusted with these PRs:
- https://github.com/mundialis/actinia-module-plugin/pull/20
- https://github.com/mundialis/actinia_satellite_plugin/pull/9
- https://github.com/mundialis/actinia-stac-plugin/pull/54
- https://github.com/mundialis/actinia_statistic_plugin/pull/13


anikaweinmann (approved):

metzm (approved):
Many substantial changes, therefore I suggest to merge rather soon and fix any issues in separate PRs.

mmacata:
I agree. I kind of wait for a fix of #332 to have a running actinia when all of these MRs are merged to be able to see directly if anything is not running smoothely.

PR #333, opened 2022-03-28
Pystac issue

joaherrerama:
Pystac dependency problem


mmacata:
I guess not all changes here belong to the pystac fix? Please suggest them in a separate MR, so it is more clear which resolves the pystac error.

PR #335, opened 2022-03-29
Pystac issue

joaherrerama:
Pystac dependency problem
cleaner version of #333


mmacata (approved):
Tested it locally and it works! Thanks.

PR #338, opened 2022-04-01
Fix wheel build in gha

mmacata:
From release 3.3.0, the build of the wheel in the github action fails due to rasterio dependencies, see [action for 3.3.0](https://github.com/mundialis/actinia_core/runs/5772793800?check_suite_focus=true) and[ action for 4.0.0](https://github.com/mundialis/actinia_core/runs/5787156855?check_suite_focus=true). This PR fixes it (still testing).

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -27,7 +27,7 @@ jobs:
       #     python-version: '3.x'
       - name: Install dependencies
         run: |
-          apt-get update && apt-get install python3 python3-pip python3-venv libgdal-dev -y
+          sudo apt-get update && apt-get install python3 python3-pip python3-venv libgdal-dev -y
           pip3 install gdal
           pip3 install rasterio==1.2.10
           # apt install git python3-venv -y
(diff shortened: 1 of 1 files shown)


mmacata:
After many tries and errors, finally solved: https://github.com/mmacata/actinia_core/runs/6081948241?check_suite_focus=true

Some pitfalls were:
- `ubuntu-latest` is still ubuntu:20.04 (locally with ubuntu:latest it did work)
- specifying an image which gha doesn't know (e.g. alpine-latest, ubuntu:latest, ...) leads to the behaviour that the job is peding until timeout without further notification
- the `actions/setup-python@v2` action behaves different than installing `python3`, `python3-pip` and `python3-venv` manually
- for the old ubuntu, `libgdal-dev` is needed to install python gdal which is needed by rasterio
- `libgdal-dev` asks questions about the users region during installation. In the workflow run, the output is not telling that installation didn't work, only later during installation of `rasterio` the job fails. Setting `DEBIAN_FRONTEND=noninteractive` helps in this case.

marcjansen:
Very nice. Thanks, please merge. Can you please squash the commits into one with a meaningful message?

marcjansen (approved):

Issue #271, opened 2021-10-25
Allow more GRASS `t.*` modules


metzm:
Currently, only a very limited set of GRASS `t.*` modules is allowed in https://github.com/mundialis/actinia_core/blob/main/src/actinia_core/core/common/config.py#L45. It would be very useful, e.g. for the openEO GRASS backend, to allow more, if not all GRASS `t.rast.*` modules by default.

@mmacata @anikaweinmann There must be a reason for this relatively short whitelist?

mmacata:
@veroandreo which modules for vector and imagery processing would you like to have included?

veroandreo:
Assuming the user is allowed to use the endpoint to download and import landsat and/or sentinel imagery, I'd add: 
i.group, i.pansharpen, i.tasscap, i.segment, i.colors.enhance, i.cluster, i.gensig, i.maxlik, i.landsat.toar, i.atcorr (these are core modules that I commonly used within grass when working with imagery).

And for vector, as I mostly use vectors to query raster data, I'd add: v.what.rast, v.rast.stats, v.what.strds, v.db.select, v.db.update, v.db.univar, v.select, v.overlay, v.patch, v.dissolve, v.clean

anikaweinmann:
The whitelist should be configurable and added to the config.
To default whitelist should added:
-  t.rast.aggregate, t.rast.aggregate.ds, t.rast.accumulate, t.rast.accdetect, t.rast.algebra, t.rast.what, t.rast.gapfill, t.rast.aggr_func,  t.create, t.register, t.list, t.remove, t.unregister
- v.what.rast, v.rast.stats, v.what.strds, v.db.select, v.db.update, v.db.univar, v.select, v.overlay, v.patch, v.dissolve, v.clean
- i.group, i.pansharpen, i.tasscap, i.segment, i.colors.enhance, i.cluster, i.gensig, i.maxlik, i.landsat.toar, i.atcorr

neteler:
> Which module is t.rast.aggr_func?

AFAIK it has been renamed to
https://github.com/mundialis/openeo-addons/tree/master/t.rast.udf

anikaweinmann:
Will be deployt with actinia-core >= 4.1.0

PR #341, opened 2022-04-22
Add more modules and configuration to add additinal modules in config

anikaweinmann:
- Add more modules to https://github.com/mundialis/actinia_core/issues/271.
- rename white list to allow list
- add configurable additinal modules via `+additional_allowed_modules` in `GRASS` section in actinia.cfg file

- update dev Dockerfile to latest actinia image (with GRASS8)

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -92,7 +92,7 @@ To lint your local changes, run
 
 Inside the container, you can run GRASS GIS with:
 ```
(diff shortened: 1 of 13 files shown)


mmacata (approved):
Thanks a lot for this!
It would be helpful to have an example in some config, e.g. actinia-core-alpine/actinia.cfg but no hinderance for me to be merged.

Issue #328, opened 2022-03-21
importer: test for new parameters resampling and resolution


juleshaas:
> You could also add a test for the new functionality

Yes, it would be good to add a test. Material may be used from here:
https://github.com/OSGeo/grass/tree/main/scripts/r.import/testsuite/

_Originally posted by @neteler in https://github.com/mundialis/actinia_core/issues/298#issuecomment-1009272006_

juleshaas:
tests added in PR #349

PR #349, opened 2022-05-06
Tests for importer: parameters resample and resolution

juleshaas:
PR

- adds test for the parameters `resample` and `resolution` of the raster importer
- does some linting in the process chains
- fixes some typos

solves #328


anikaweinmann (approved):
Thank you, looks very good :)

PR #354, opened 2022-06-15
Fix STRDS removement and unregistration of rasters in STRDS

anikaweinmann:
see title

### DIFF
--- a/src/actinia_core/processing/actinia_processing/ephemeral/persistent_processing.py
+++ b/src/actinia_core/processing/actinia_processing/ephemeral/persistent_processing.py
@@ -350,12 +350,14 @@ class PersistentProcessing(EphemeralProcessing):
                             f"SELECT * FROM dba.{table}")
                 continue
             # for example raster_register_xxx tables are not in both dbs
-            if table not in table_names1:
-                con.execute(f"CREATE TABLE {table} AS "
-                            f"SELECT * FROM dba.{table}")
-                continue
-            combine = f"INSERT OR IGNORE INTO {table} SELECT * FROM dba.{table}"
-            con.execute(combine)
+            if table in table_names1:
+                # drop table so that raster can also be unregistered
+                con.execute(f"DROP TABLE {table}")
+            con.execute(f"CREATE TABLE {table} AS "
+                        f"SELECT * FROM dba.{table}")
+        for table in table_names1:
+            if table not in table_names2:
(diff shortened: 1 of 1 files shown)


src/actinia_core/processing/actinia_processing/ephemeral/persistent_processing.py
```diff
@@ -433,19 +443,31 @@ def _merge_mapset_into_target(self, source_mapset, target_mapset):
...
-                (stdout_buff, stderr_buff) = p.communicate()
-                if p.returncode != 0:
-                    raise AsyncProcessError(
-                        "Unable to merge mapsets. Error in linking:"
-                        " stdout: %s stderr: %s" % (stdout_buff, stderr_buff))
+                self._copy_folder(
+                    source_path, target_path,
+                    msg="merge mapsets. Error in linking")
+
+    def _copy_folder(
```
metzm, line 450: Why not using `shutil.copytree()`?

src/actinia_core/processing/actinia_processing/ephemeral/persistent_processing.py
```diff
@@ -324,42 +324,59 @@ def _update_views_in_tgis(self, tgis_db_path):
...
+            f"PRAGMA table_info({table_name})") if row[-1] == 1]
+        if len(selection) == 0:
+            primary_key = ""
+        else:
+            primary_key = selection[0]
+
+        for col in columns:
+            if col not in skip_columns:
+                update_statement = f"UPDATE {table_name} SET {col} = " \
+                    f"REPLACE({col}, '{source_mapset}', '{target_mapset}')"
```
metzm, line 344: What if the name of the map and of the source mapset are the same, e.g. `elevation@elevation`? This would then not only replace the mapset name, but also the map name?

metzm:
What about creating a list of space-time datasets to be copied and using `t.copy`?

src/actinia_core/processing/actinia_processing/ephemeral/persistent_processing.py
marcjansen, line 344:
```suggestion
                    f"REPLACE({col}, '@{source_mapset}', '@{target_mapset}')"
```

untested

metzm: Works for the column "id" but not for the column "mapset".

PR #356, opened 2022-07-04
Add dependencies for actinia-parallel-plugin

anikaweinmann:
Add dependencies for https://github.com/mundialis/actinia-parallel-plugin/pull/1


mmacata (approved):
Thanks!

PR #357, opened 2022-07-11
Endpoint configuration

anikaweinmann:
With this change it is possible to configure the endpoints (including methods) of actinia-core.

You can add a endpoints configuration csv file with the following structure which list all desired endpoints including method:
  ```
  Class_of_the_endpoint;method1,method2
  Class_of_the_endpoint2;method1
  ```
Add the path of the csv file to the actinia.cfg as `API.endpoints_config` entry.


src/actinia_core/core/common/endpoint_config.py
```diff
@@ -0,0 +1,87 @@
+# -*- coding: utf-8 -*-
```
mmacata, line 1: The two methods of this module are only used in all the endpoint classes, right? Then I would move this maybe to `src/actinia_core/rest/base/` ?

mmacata (commented):
Thanks! In general I approve, only maybe the endpoint_config module could be moved?

Issue #359, opened 2022-07-20
Vector info bug


anikaweinmann:
If a vector has no database connection because it has no tabel, actinia returns an error by requesting the information of the vector map:
```
# test_random_vect created with e.g. grass grassdb/nc_spm_08/PERMANENT/ --exec v.random output=test_random_vect npoints=10

curl -u ${AUTH} ${URL}/locations/nc_spm_08/mapsets/PERMANENT/vector_layers/test_random_vect | jq
{
    ....
    {
      "executable": "v.info",
      "id": "2",
      "mapset_size": 375,
      "parameter": [
        "map=test_random_vect@PERMANENT",
        "-h"
      ],
      "return_code": 0,
      "run_time": 0.10043096542358398,
      "stderr": [
        ""
      ],
      "stdout": "COMMAND: v.random output=\"test_random_vect\" npoints=10 layer=\"-1\" zmin=0.0 zmax=0.0 column_type=\"double precision\"\nGISDBASE: /home/user/actinia/grassdb\nLOCATION: nc_spm_08 MAPSET: PERMANENT USER: user DATE: Wed Jul 20 12:05:35 2022\n"
    },
    {
      "executable": "v.info",
      "id": "3",
      "mapset_size": 375,
      "parameter": [
        "map=test_random_vect@PERMANENT",
        "-c"
      ],
      "return_code": 1,
      "run_time": 0.10044598579406738,
      "stderr": [
        "ERROR: Database connection for map <test_random_vect@PERMANENT> is not defined in DB file",
        ""
      ],
      "stdout": ""
    }
  ],
  "progress": {
    "num_of_steps": 3,
    "step": 3
  },
  "resource_id": "resource_id-9feaee6d-8abf-4253-8437-6a5ee2556d0d",
  "status": "error",
  "time_delta": 0.4676177501678467,
  "timestamp": 1658318803.6137428,
  "urls": {
    "resources": [],
    "status": "http://...:8088/api/v3/resources/actinia-gdi/resource_id-9feaee6d-8abf-4253-8437-6a5ee2556d0d"
  },
  "user_id": "actinia-gdi"
}

```
The endpoint should not return an error if no table is set for the vector map.

linakrisztian:
New pull request: #408 (old one #364 )

linakrisztian:
tbd:
- fix test-failure (see todos in #408 )
- update documentation (https://github.com/actinia-org/actinia-api/pull/13)

PR #364, opened 2022-07-28
Fix Vector info bug #359

linakrisztian:
#359: By changing v.info -c to v.db.connect -c, GRASS does not return an error for vector maps with missing table. Instead this message is printed: "Map <test_random_vect> is not connected to a database". Thus the endpoint does not return an error anymore.


griembauer (approved):
Thanks, looks good and tested successfully locally!

src/actinia_core/processing/actinia_processing/ephemeral/vector_layer.py
```diff
@@ -65,7 +65,7 @@ def _execute(self):
             "flags": "h"}
 
         pc["3"] = {
-            "module": "v.info",
+            "module": "v.db.connect",
             "inputs": {"map": vector_name + "@" + self.mapset_name},
             "flags": "c"}
```
metzm, line 70: Do you really want to print types/names of table columns for the specified layer (default is layer 1) or rather print all map connection parameters for all layers, would be empty if there are no connections?

```suggestion
            "flags": "g"}
```

linakrisztian: @anikaweinmann or @mmacata : The old case (v.info -c) also printed only type/names of the specified layer (default layer 1). Should we change it as @metzm suggested?

metzm:
A GRASS vector can have no database connection at all or one or several database connections, one possible connection for each layer. The common case is that a vector has one database connection, the corresponding layer is layer 1. For other cases `v.info -c` or `v.db.connect -c` does not make sense because no columns would be reported or only columns for layer 1, ignoring attribute columns for other layers.
The class `EphemeralVectorInfo()`reports general information about a vector. Therefore I would argue that this is the wrong place for column names and types of a specific layer. Instead, any database connections as reported by `v.db.connect -g()` would be useful information in this place. After that, a user might want to know column names and types of a specific vector layer, see also https://github.com/OSGeo/grass/blob/main/python/grass/script/vector.py#L101.

Maybe introduce a new endpoint where the layer of a vector can be specified?

metzm:
Instead of a new endpoint for a vector layer, the syntax also recognized by TGIS could be used: `<name>:<layer>@<mapset>`. If a layer is given, `v.db.connect -c layer=<layer> map=<name>` could be used, otherwise `v. db.connect -g map=<name>`.

src/actinia_core/processing/actinia_processing/ephemeral/vector_layer.py
anikaweinmann: I think `v.db.connect -c` is the one which is like it is now. So I would prefer this. The style of `g` is anonther one than we have a breaking change, or not?

anikaweinmann: Check with `v.db.connect -g` if there are more than 1 layer/tables or no layers (count number of rows). If there is only one layer the output should be `v.info -c ` as before, by multiple layer/tables `v.db.connect -g` should be inside the results. By no layer/table nothing has to be added to the results.

Open question: how to write this in the api docs.

metzm (approved):

PR #371, opened 2022-08-01
actinia tests: fix test descriptions

neteler:
This PR fixes some test descriptions (wording) and a few typos.
Hope I got them right.

### DIFF
--- a/tests/test_async_processing_import_export.py
+++ b/tests/test_async_processing_import_export.py
@@ -552,7 +552,7 @@ class AsyncProcessTestCase(ActiniaResourceTestCaseBase):
     @unittest.skipIf(no_stac_plugin, "STAC Plugin not installed")
     def test_stac_import(self):
         """
-            Code test STAC collection importation with http reponse 200
+            Test of STAC collection import with http response 200
         """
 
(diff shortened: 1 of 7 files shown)


anikaweinmann (approved):

tests/test_async_processing_import_export.py
```diff
@@ -598,7 +598,7 @@ def test_stac_source_filter_error_import(self):
 
     def test_raster_import_resample_resolution(self):
         """
-        Code test raster import with set resampling and resolution with http reponse 200
+        Test raster import by setting resampling and resolution, with http response 200
```
juleshaas, line 601: I'd rather keep the `with set resampling an resolution` here and in the following tests. Maybe change it to `defined resampling and resolution`, to make clear these values are set/defined during the import?

neteler: OK, I have reworded it in 73839ed.

juleshaas (approved):

tests/test_async_processing_import_export.py
juleshaas: Looks good to me.

PR #376, opened 2022-08-09
Allow separate config for worker

mmacata:
Currently, the processing part in actinia reads the config from the process queue in redis (when configured) via the `resource data container` object. This means that the config is used which is meant for the actinia which receives the job and writes it into the queue.
The worker config can only be specified for

- REDIS_QUEUE_SERVER_URL
- REDIS_QUEUE_SERVER_PORT
- REDIS_QUEUE_SERVER_PASSWORD
- LOG_INTERFACE
- LOG_FLUENT_HOST
- LOG_FLUENT_PORT
- WORKER_LOGFILE

But many other config is not read from the worker config, e.g. additional redis config (for users etc), data mounts, ... which obviously can differ between the actinia instance receiving the job and the actinia worker.

This PR fixes this.
Anything I might oversee @anikaweinmann @metzm ?

### DIFF
--- a/src/actinia_core/processing/actinia_processing/ephemeral_processing.py
+++ b/src/actinia_core/processing/actinia_processing/ephemeral_processing.py
@@ -40,6 +40,7 @@ import uuid
 from flask import json
 from requests.auth import HTTPBasicAuth
 
+from actinia_core.core.common.config import global_config, DEFAULT_CONFIG_PATH
 from actinia_core.core.common.process_object import Process
 from actinia_core.core.grass_init import GrassInitializer
 from actinia_core.core.messages_logger import MessageLogger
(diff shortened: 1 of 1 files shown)


src/actinia_core/processing/actinia_processing/ephemeral/download_cache_management.py
```diff
@@ -59,37 +58,30 @@ def __init__(self, *args):
     def _execute(self):
         self._setup()
 
-        if os.path.exists(self.user_download_cache_path) and os.path.isdir(
-            self.user_download_cache_path
-        ):
-            executable = "/usr/bin/du"
-            args = ["-sb", self.user_download_cache_path]
+        if not os.path.exists(self.user_download_cache_path):
```
anikaweinmann, line 61: `os.path.exists` checks only if the path exists, but not if it is a directory or not.

mmacata: True. If there is no directory and no file it will create the folder. Do you think I should check if it exists if it is a file and then raise an error?

anikaweinmann: yes. Otherwise the next step would fail, or not?

mmacata: No it would be fine:
```
/usr/bin/du -sb download_cache_management.py 
2838	download_cache_management.py
```
Although it would indicate a config error. But an error that the download cache is not configured correctly wouldn't help a user much at this point. If it is important to check it would make more sense somewhere else, e.g. on actinia startup where the person installing actinia would be made aware of it. Do you think that should be added here in this PR? Or new feature request?

mmacata:
As for this PR all tests pass and also processing for actinia worker with same config is successful, I will merge now. Follow-Up PR is coming.

PR #379, opened 2022-09-15
Fix redoc rendering

ninsbl:
The API docs give only an error currently. Adding `&nocors` aparently helps. (maybe browser dependent?)
See: https://github.com/Redocly/redoc/issues/2009 for background
Should probably be changed other places as well. I can have a look at it if the issue is not only occurring om my browser...


mmacata:
Very cool, thanks!

PR #380, opened 2022-09-21
Add job queue name to response

mmacata:
When sending a job to actinia and `queue_type` is configured to create a new queue for each job, it would be nice to know which queue the job is written to, especially when a worker is started dedicately for this job. This PR suggest to add it in the following way:
```
{
  "accept_datetime": "2022-09-21 15:41:08.301880",
  "accept_timestamp": 1663774868.3018744,
  "api_info": {
    "endpoint": "asyncephemeralexportresource",
    "method": "POST",
    "path": "/api/v3/locations/nc_spm_08/processing_async_export",
    "request_url": "http://127.0.0.1:8088/api/v3/locations/nc_spm_08/processing_async_export"
  },
  "message": "Resource accepted",
  "queue": "job_queue_resource_id-6c6f8e20-853d-42e3-bb42-7b0e287700f1",
  "resource_id": "resource_id-6c6f8e20-853d-42e3-bb42-7b0e287700f1",
  "status": "accepted",
...
}
```
There are three options for the queue: `local`, `redis` and `per_job`.

The queue name is not only added to the response but added to the rdc in redis database. Actinia first does a preprocessing step which writes initial information into the db. For the types `local` and `per_job` the queue name can be determined correctly and is written to redis. For type `redis`, the name is build at a later step and can only be estimated which is done by setting it to e.g. `job_queue_count` (composited of the `global_config.WORKER_QUEUE_PREFIX` and the word `count` - suggestions welcome) and as soon as the job is `running`, it is replaced with the real queue name, e.g. `job_queue_0`. The other two job types `local` and `per_job` are also replaced by the real value created in the `enqueue_job` method, but as they can be estimated well, they do not change after.

This might not be ideal for job type `redis` but as this information is mostly crucial for job type `per_job` where a new worker is started listening to a specific queue, it is not really a dealbreaker in my opinion.


anikaweinmann (approved):

src/actinia_core/models/response_models.py
```diff
@@ -327,6 +327,10 @@ class ProcessingResponseModel(Schema):
             'type': 'string',
             'description': 'The unique resource id'
         },
+        'queue': {
+            'type': 'string',
+            'description': 'The name of the queue where the job is queued'
```
neteler, line 332:
```suggestion
            'description': 'The name of the queue in which the job is queued'
```

PR #381, opened 2022-09-23
Add possibility to use keycloak authentication token

anikaweinmann:
This merge requests allows to use a keycloak token for authentication, where the user attributes, role and group have to be included.

The changes includes:
- creation of a `ActiniaUserBase` where the original `ActiniaUser` inherit from
- Adding a new `ActiniaKeycloakUser` class which is used when it is configured in the config file via:
```

```
If the keycloak authentication token is configured the basic authenication will not work at the same time.


src/actinia_core/core/common/app.py
```diff
@@ -136,10 +140,15 @@
...
-flask_api._swagger_object["securityDefinitions"] = {
-    "basicAuth": {"type": "basic"}
-}
-flask_api._swagger_object["security"] = [{"basicAuth": []}]
-
-auth = HTTPBasicAuth()
+# authentication method
+global_config.read(DEFAULT_CONFIG_PATH)
+if global_config.KEYCLOAK_CONFIG_PATH:
+    auth = HTTPTokenAuth(scheme="Bearer")
```
mmacata, line 146: Maybe it is possible to add flask description for bearer auth as well? (https://swagger.io/specification/v2/#securityDefinitionsObject)

src/actinia_core/core/common/config.py
```diff
@@ -708,6 +736,26 @@ def read(self, path=DEFAULT_CONFIG_PATH):
...
+                if config.has_option("KEYCLOAK", "CONFIG_PATH"):
+                    keycloak_cfg_path = config.get("KEYCLOAK", "CONFIG_PATH")
+                    if os.path.isfile(keycloak_cfg_path):
+                        self.KEYCLOAK_CONFIG_PATH = keycloak_cfg_path
+                    # else:
+                    #     log.warning(
+                    #         f"Keycloak is configured, but configfile is not "
+                    #         "an existing file! Using Redis for user "
+                    #         "management."
+                    #     )
```
mmacata, line 749: Is this supposed to be activated later? If yes, please add a TODO, if not, please remove.

src/actinia_core/core/common/keycloak_user.py
```diff
@@ -0,0 +1,323 @@
+# -*- coding: utf-8 -*-
```
mmacata, line 1: Please move file one level up, out of `common` folder. This folder was meant for modules / classes which are used by actinia plugins as well, although this wasn't communicated well enough ;)

src/actinia_core/core/common/config.py
```diff
@@ -319,6 +322,14 @@ def __init__(self):
...
         self.ENDPOINTS_CONFIG = None
 
+        """
+        KEYCLOAK: has only to be set if keycloak server is configured with
+        actinia client and actinia attributes for the users
+        """
+        self.KEYCLOAK_CONFIG_PATH = None
+        self.KEYCLOAK_GROUP_PREFIX = None
+        self.KEYCLOAK_ATTR_PREFIX = None
+
```
mmacata, line 332:
```suggestion
        """
        # Json file generated by keycloak with configuration,
        # e.g. /etc/default/keycloak.json
        self.KEYCLOAK_CONFIG_PATH = None
        # e.g. /actinia-user/
        self.KEYCLOAK_GROUP_PREFIX = None
        # Prefix to distinguish parameters inside keycloak from parameters
        # used by other applications, e.g. actinia
        self.KEYCLOAK_ATTR_PREFIX = None

```

src/actinia_core/core/common/keycloak_user.py
```diff
@@ -0,0 +1,323 @@
...
+                keycloak_cfg = json_load(f)
+                KEYCLOAK_URL = keycloak_cfg["auth-server-url"]
+                REALM = keycloak_cfg["realm"]
+                CLIENT_ID = keycloak_cfg["resource"]
+                CLIENT_SECRET_KEY = keycloak_cfg["credentials"]["secret"]
+        else:
+            raise Exception(
+                "KEYCLOAK_CONFIG_PATH is not a valid keycloak configuration "
+                "for actinia"
+            )
```
mmacata, line 62: Not really sure about this suggestion, but somehow I find it confusing that config is stored / read in different places. I know that this is only the config read directly from the json generated by keycloak, but maybe it would be more nice to have this section in the config.py, storing the values directly to `global_config` object instead of working with global variables here!?

mmacata (commented):
Very nice, thank you for this long-awaited feature :dancing_women: :tada: 

Besides some small typos I have some comments, none of them are really blocking, but I would like to know your opinion about them.

src/actinia_core/core/common/config.py
anikaweinmann: It is not possible to activate it because then I get an circle of imports. But I would find it good if this would be logged somewhere. Ideas are very welcome

src/actinia_core/core/common/user.py
```diff
@@ -34,29 +34,19 @@
...
 from itsdangerous import JSONWebSignatureSerializer
-from actinia_core.core.common.config import global_config
-from actinia_core.core.redis_user import redis_user_interface
 
-__author__ = "Sören Gebbert"
-__copyright__ = (
-    "Copyright 2016-2018, Sören Gebbert and mundialis GmbH & Co. KG"
+from actinia_core.core.common.user_base import (
+    ActiniaUserBase,
 )
```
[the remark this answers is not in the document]
anikaweinmann: The class of the user has not been moved. So no action should be needed. And ActiniaUserError is not in the list of dependencies of the plugins.

src/actinia_core/core/common/keycloak_user.py
anikaweinmann: yes I can do it

src/actinia_core/core/common/keycloak_user.py
anikaweinmann: But two modules (module-plugin and parallel-plugin) are using ActiniaUser. So maybe they should also include keycloak and use this new user class.

src/actinia_core/core/common/config.py
mmacata: Maybe use a simple `print` statement? Not so nice as it cannot be configured if it is json or not, but maybe better than nothing?

src/actinia_core/core/common/keycloak_user.py
mmacata: alright then.

mmacata:
:woman_dancing:

PR #388, opened 2022-10-21
Add initialisation of self.queue

### LINKED ISSUE #384: Bug: GET /grass_modules
The GET */grass_modules* route is no longer behaving as expected after updating [actinia_core](https://github.com/mundialis/actinia_core) from v4.2.1 to v4.3.0 and [actinia_module_plugin](https://github.com/mundialis/actinia-module-plugin) from v2.3.1 to v2.4.0. The route now hangs leaving a temporary location, such as, `location_for_listing_modules_4ac35907-cc15-4832-a945-d5459c966f9d` as an artifact.

From the light debugging I've done so far, the issue does not seem to stem from the **actinia_module_plugin (v2.4.0)** because if I only update the actinia_module_plugin to v2.4.0 and leave actinia_core at v4.2.1, everything works as expected. So, it appears like something in the **actinia_core (v4.3.0)** is causing the issue, maybe with modifications to EphemeralProcessing (#380).

Let me know if I can provide you with anything else to help debug.

juleshaas:
PR fixes endpoint `GET /grass_modules`

closes #384

### DIFF
--- a/src/actinia_core/core/resource_data_container.py
+++ b/src/actinia_core/core/resource_data_container.py
@@ -112,6 +112,7 @@ class ResourceDataContainer(object):
         self.orig_datetime = orig_datetime
         self.user_data = None
         self.storage_model = "file"
+        self.queue = None
 
     # def __str__(self):
     #    return str(self.__dict__)


mmacata (approved):
Nice thanks!

PR #390, opened 2022-10-24
Start with the adjustment of the interim results

anikaweinmann:
Actinia should be able to be configured so that the interim results can also only be saved if an error has occurred during processing and the interim results should be deleted after successful job resumption.

How to configure the saving of interim results and how to adjust the configuration for the used endpoints is documented in the [job_resumption.md](https://github.com/anikaweinmann/actinia_core/blob/interim_results_on_error/job_resumption.md).

Adjustment of API doc in https://github.com/mundialis/actinia-api/pull/11.


job_resumption.md
```diff
@@ -0,0 +1,470 @@
...
+            ],
+            "flags": "g",
+            "stdout": {"id": "r_info_1", "format": "kv", "delimiter": "="}
+        }
+    ],
+    "version": "1"
+}
+```
+
+`pc_template_forput.json`: complete process chain without template:
```
mmacata, line 417:
```suggestion
`pc_template_forput.json`: complete process chain without template (can be found in actinia response) :
```

src/actinia_core/core/common/config.py
```diff
@@ -774,9 +792,27 @@ def read(self, path=DEFAULT_CONFIG_PATH):
...
-                    self.SAVE_INTERIM_RESULTS = config.getboolean(
-                        "MISC", "SAVE_INTERIM_RESULTS"
-                    )
+                    save_interim = config.get("MISC", "SAVE_INTERIM_RESULTS")
+                    if save_interim == "True":
+                        self.SAVE_INTERIM_RESULTS = True
+                    elif save_interim == "False":
+                        self.SAVE_INTERIM_RESULTS = False
+                    else:
+                        self.SAVE_INTERIM_RESULTS = save_interim
```
mmacata, line 801: I find it a bit confusing that a config flag has different options which are differen data types. Would it be better to have maybe `onError`, `always` and `never` ? This way, all options are strings. But then it is not backward compatible anymore, so open for discussions here.

mmacata: As discussed, leave as is.

mmacata (approved):

PR #391, opened 2022-10-25
Create temperature.py

DollyMehra:


mmacata:
Thanks for your work, but this script has nothing to do with actinia and is not enhancing it's functionality, therefore doesn't fit into this repository.

DollyMehra:
ok thanks!!!!!!!!!!!!!

PR #392, opened 2022-10-31
Resource storage older than X days deletion

anikaweinmann:
Add parameter for the resource_storage endpoint to delete only resource files older than x days, e.g.:

```
curl -X DELETE -u XXX:XXX http://127.0.0.1:8088/api/v3/resource_storage?olderthan=5
```
The API is adjusted in https://github.com/mundialis/actinia-api/pull/10.


mmacata (approved):
Thanks!

Issue #398, opened 2022-12-14
Storing and accessing credentials (e.g. database logins) as actinia user


ninsbl:
Is there already a way to securely store and access e.g. database credentials as an actinia user?

When data is to be fetched from or exported to external databases, one needs some form of storing login credentials for actinia users.
Maybe also credential for data portals, if they are supposed to be accessed userspecific...

It does not seem right to include credentials in API calls in plain text.

One option I could imagine is that a user may upload an encrypted [keepass](https://keepass.info/) file that in itself is secured (and later accessed) with the actinia password. In actinia credentials could be fetched with e.g. https://pypi.org/project/pykeepass/

But I am no security expert and maybe that is also a terrible idea with regards to security?

ninsbl:
Now I wrote a function to transfer entries in a KeePass database file into environment variables, that again could be used to authenticate users at data sources services / databases.

```
import os
def keepass_to_env(
    keepass_file, keepass_pwd, title, username_var, password_var, first=True
):
    from pykeepass import PyKeePass

    kp = PyKeePass(keepass_file, password=keepass_pwd)
    entry = kp.find_entries(title=title, first=first)
    os.environ[username_var] = entry.username
    os.environ[password_var] = entry.password
    return None
``` 

For use in actinia modules, two questions remain:
1) how to securely transfer and store the KeePass file to actinia / user data area and
2) how to unlock the KeePass file with the actinia password (assuming that it is used to protect the KeePass file)...

Another approach could be bitwarden (that could be connected to KeyCloak if I understood correctly)...

ninsbl:
Apache Airflow (also a Python based web app) supports different external backends, like e.g. the [Hashicorp Vault](https://www.vaultproject.io/) for [secrets management](https://github.com/apache/airflow/tree/main/airflow/secrets).
See: https://github.com/apache/airflow/tree/main/airflow/providers/hashicorp/secrets for inspiration on how secrets backends are implemented in Airflow.
_Hashicorp Vault_ is an Open Source project and could even be bundled with actinia (at least in theory). It also powers secrets handling in GitHub actions. Would be cool if secrets could be stored in _Vault_ and then accessed like in GitHub actions with a specific kind of string (`${{ secrets.SENTINELHUB_AUTH_TOKEN }}`).

Without external secrets backends, Airflow stores credentials as encrypted strings in a _PostgreSQL_ database. Such a solution could possibly be a first approach for actinia too...

Just collecting some ideas....

[...]

PR #411, opened 2023-01-24
document auth property in webhooks section of processing chain

ninsbl:
This PR is an attempt to document the `auth` property in the `webhooks` section of a processing chain. Currently, this is not visible in swagger.json / ReDoc.
I have not tested it and I am not sure this is the right way to add to the documentaion...

It would be good it `auth` would support _base64_ encoded strings (instead of username and password in plain text), or even better, if actinia had secrets management (like e.g. in github actions, see also: #398).

### DIFF
--- a/src/actinia_core/models/process_chain.py
+++ b/src/actinia_core/models/process_chain.py
@@ -466,6 +466,11 @@ class Webhooks(Schema):
     """
     type = 'object'
     properties = {
+        'auth': {'type': 'string',
+                 'description': 'Authentication information (username and password) '
+					            'to authenticate to the webhook url.Given as '
+					            '"username:password" string. The username and '
+					            'password should not contain a ":".'},
         'update': {'type': 'string',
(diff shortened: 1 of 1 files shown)


src/actinia_core/models/process_chain.py
```diff
@@ -466,6 +466,11 @@ class Webhooks(Schema):
     """
     type = 'object'
     properties = {
+        'auth': {'type': 'string',
+                 'description': 'Authentication information (username and password) '
+					            'to authenticate to the webhook url.Given as '
```
mmacata, line 471:
```suggestion
					            'to authenticate to the webhook url. Given as '
```

mmacata (approved):
Cool, thanks!
This should be the corret place, yes. To test it is possible to start actinia locally and then call `/api/v3/swagger.json` to receive the json. This URL can be entered into https://petstore.swagger.io/ - for me it doesn't work in any browser as a local URL is adressed but it used to work in Firefox.

mmacata (approved):

PR #424, opened 2023-05-04
docker CI: only run CI in actinia-org remote

neteler:
Avoid docker CI runs in forks as being useless and a waste of CPU power/energy.


mmacata (approved):
merci :)

PR #443, opened 2023-06-05
Increase allowed length of additional version info key

mmacata:
To allow information for `actinia_docker_version` which has 22 characters, it is necessary to allow more than 20 chars.


mmacata:
As this is a really tiny change, I will merge immediately.

src/actinia_core/version.py
```diff
@@ -91,7 +91,7 @@ def valid_additional_version_info_key(cand):
     a length between 1 and 20 and must not be any of the reserved keys which
```
marcjansen, line 91: Should be adjusted in the comment as well

PR #446, opened 2023-06-07
Update grass docker image reference

mmacata:
Update docker image source from `mundiais/grass-py3-pdal` to `osgeo/grass-gis`


mmacata:
Tested locally, as this is only a tiny change, I will merge now.

PR #447, opened 2023-06-07
Change runtime if runtime is higher 2147483647

anikaweinmann:
In the worker the allowed `job_timeout` of an user is not allowed to be higher than 2147483647.
For this the `job_timeout` is set to `-1` which means never expire if the user has a timout higher.


mmacata (approved):
Cool, thank you! I would just rephrase the one comment a bit, else pls merge!

PR #455, opened 2023-06-22
Fixes currently failing test

mmacata:
This PR
- fixes `test_async_processing_new` by catching an additional exception of werkzeug
- Calls `./tests_with_redis.sh` via `sh` to not depend on `chmod` to have local test setup clean
- do not use `requirements.txt` anymore but install dependencies from `pyproject.toml`

### DIFF
--- a/src/actinia_core/cli/webhook_server.py
+++ b/src/actinia_core/cli/webhook_server.py
@@ -49,9 +49,9 @@ def finished():
         pprint(json.loads(request.get_json()))
     except BadRequest:
         pass
-    except TypeError:
-        pass
     except UnsupportedMediaType:
+        pass
+    except TypeError:
         pass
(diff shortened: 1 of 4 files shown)


src/actinia_core/cli/webhook_server.py
```diff
@@ -61,6 +64,8 @@ def update():
         pass
     except TypeError:
```
anikaweinmann, line 65: Not quite sure, but to me `TypeError` sounds more general than `UnsupportedMediaType`, so I would have expected the order of the two to be the other way around.

anikaweinmann (approved):
Thank you.

src/actinia_core/cli/webhook_server.py
mmacata: I changed it.

PR #456, opened 2023-06-26
Make resource storage delete asynchron

anikaweinmann:
Make resource storage delete asynchron because of timeout when the resource storage has a large amount of data.


mmacata (approved):
Thanks for these enhancements!

PR #458, opened 2023-06-27
Black

anikaweinmann:
Changes for new black version (22.3.0 to 23.3.0)


mmacata (approved):
Nice, thank you. In general I agree, just one question:
Did the allowed line length change? It looks like some lines which were ok before were shortened in this PR.

PR #459, opened 2023-06-29
fix HOME env

anikaweinmann:
The HOME environment variable was set to `/tmp/:/root` to fix this the `env` attributes are now subdivided into:

1. variables where use self.env and enviroment variable values (used for `["PATH", "PYTHONPATH"]`)
2. use self.env value for the rest (`HOME` is before set to $HOME if $HOME is set)

resulting in: 
```
GISBASE=/usr/local/grass
GISRC=/tmp/tmplc_46gr7/gisrc
LD_LIBRARY_PATH=/usr/local/grass/lib
GRASS_ADDON_PATH=/root/.grass8/addons/
GRASS_VERSION=7.7.svn
PYTHONPATH=:/usr/local/grass/etc/python
GRASS_MESSAGE_FORMAT=plain
GRASS_SKIP_MAPSET_OWNER_CHECK=1
GRASS_TGIS_RAISE_ON_ERROR=1
GIS_LOCK=48
HOME=/root
GRASS_ADDON_BASE=/root/.grass8/addons/
PATH=/usr/local/grass/bin:/usr/local/grass/scripts:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```


src/actinia_core/core/grass_init.py
```diff
@@ -36,7 +36,8 @@
 __license__ = "GPLv3"
 __author__ = "Sören Gebbert"
 __copyright__ = (
-    "Copyright 2016-2018, Sören Gebbert and mundialis GmbH & Co. KG"
+    "Copyright 2016-2023, Sören Gebbert, Anika Weinmann and mundialis GmbH & "
+    "Co. KG"
 )
 __maintainer__ = "Sören Gebbert"
 __email__ = "soerengebbert@googlemail.com"
```
mmacata, line 43:
```suggestion
__maintainer__ = "mundialis"
```

src/actinia_core/core/grass_init.py
```diff
@@ -113,7 +114,6 @@ def set_grass_environment(
 
         """
         self.env["GIS_LOCK"] = str(os.getpid())
-        self.env["HOME"] = "/tmp/"
```
mmacata, line 116: Interesting. I still wonder why it worked for the webserver and not for the worker.

mmacata (approved):

PR #466, opened 2023-08-24
Use reusable pre-commit hooks for linting

mmacata:
from https://github.com/mundialis/github-workflows


anikaweinmann (approved):
Thank you

PR #472, opened 2023-09-02
mdformat

mmacata:
This PR adds the new `mdformat` linting workflow from mundialis/github-workflows.
For pre-commit and linting, it also reformats all md files.


neteler (approved):
Thanks for the reformatting!

PR #473, opened 2023-09-03
Add pre-commit to renovate config

mmacata:
opt-in to update pre-commit hooks in renovate config


mmacata:
As already tested in mundialis/v.example repository, I will merge directly.

PR #485, opened 2023-10-16
Enable import via vsicurl

mmacata:
See previous discussions in https://github.com/actinia-org/actinia-core/pull/482


mmacata:
Developments in actinia-org organisation was not needed, so back to using https://github.com/actinia-org/actinia-core/pull/482

PR #491, opened 2023-10-26
Checking pixellimit for r.import commands

linakrisztian:
Check the current `r.import` command against the user cell limit (within `ephemeral_processing.py`) + added tests for this check

PR additionally contains small linting changes:
- `test_download_cache.py` (line too long)
- format markdown

### DIFF
--- a/src/actinia_core/processing/actinia_processing/ephemeral_processing.py
+++ b/src/actinia_core/processing/actinia_processing/ephemeral_processing.py
@@ -1196,6 +1196,45 @@ class EphemeralProcessing(object):
         ):
             shutil.rmtree(self.temp_grass_data_base, ignore_errors=True)
 
+    def _check_pixellimit_rimport(self):
+        """Check the current r.import command against the user cell limit.
+
+        Adjust the r.import command to a meaningful state
+        so that the user cell limit is not reached and the mapset can be
+        accessed again.
+
+        Raises:
+            This method will raise an AsyncProcessError exception
+
+        """
+        # TODO: define extent_region if r.import extent=region
+        extent_region = None
+
+        # build VRT with gdalbuildvrt
+        if extent_region:
+            # first query region extents
+            errorid, stdout_buff, stderr_buff = self.ginit.run_module(
+                "g.region", ["-ug"]
+            )
+
+            if errorid != 0:
+                raise AsyncProcessError(
+                    "Unable to check the computational region size"
+                )
+            # TODO: build vrt with gdalbuildvrt and -te flag
+            # (extents parsed from previous command)
+        else:
+            None
+            # TODO: build vrt with gdalbuildvrt (for whole input)
+        # TODO:
+        # query raster size with gdalinfo
+        # if resample and resoltuion set:
+        #   include to expected raster output
+
+        # compare estimated raster output size with pixel limit
+        # and raise exception if exceeded with message
+        # set e.g. region smaller
+
     def _check_reset_region(self):
         """Check the current region settings against the user cell limit.
 
@@ -1470,6 +1509,14 @@ class EphemeralProcessing(object):
                 str(process.executable_params),
             )
             self._send_resource_update(message)
+
+        # Check pixel limit for r.import operations
+        # TODO: test if by that all r.imports are detected
+        if (
+            self.last_module == "r.import"
+            and process.skip_permission_check is False
+        ):
+            self._check_pixellimit_rimport()
 
         # Check reset region if a g.region call was present in the process
         # chain. By default the initial value of last_module is "g.region" to
--- a/src/actinia_core/processing/actinia_processing/ephemeral_processing.py
+++ b/src/actinia_core/processing/actinia_processing/ephemeral_processing.py
@@ -1204,44 +1204,126 @@ class EphemeralProcessing(object):
         ):
             shutil.rmtree(self.temp_grass_data_base, ignore_errors=True)
 
-    def _check_pixellimit_rimport(self):
+    def _check_pixellimit_rimport(self, process_executable_params):
         """Check the current r.import command against the user cell limit.
-
-        Adjust the r.import command to a meaningful state
-        so that the user cell limit is not reached and the mapset can be
-        accessed again.
 
         Raises:
             This method will raise an AsyncProcessError exception
 
         """
-        # TODO: define extent_region if r.import extent=region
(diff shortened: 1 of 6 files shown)


neteler:
May I suggest adding a test `test_download_commands_vsicurl` similar to

https://github.com/actinia-org/actinia-core/blob/852f159696b93848e8a7c79f3bbcecb667e02b09/tests/test_geodata_import.py#L170

The test may then execute two imports:
- `r.import` on /vsicurl/https://data.mundialis.de/geodata/lulc-germany/classification_2021/classification_map_germany_2021_v01_COG.tif --> should **fail** as too many pixels
- `r.import` on /vsicurl/https://github.com/OSGeo/grass/raw/main/raster/r.in.gdal/testsuite/data/elevation.tif  --> should **succeed** as only 150x135 pixels

linakrisztian:
@neteler Good hint, I will add a test and thanks for the example data.
 I had a look and I think it is not reasonable to put it into [actinia-core/tests/test_geodata_import.py](https://github.com/actinia-org/actinia-core/blob/852f159696b93848e8a7c79f3bbcecb667e02b09/tests/test_geodata_import.py#L170). Within this test the download and import for vector and raster data is tested.
I will create some separate test-file `test_raster_import_pixellimit.py`, checking the pixellimit for all `r.import` commands.

tests/test_raster_import_pixellimit.py
```diff
@@ -0,0 +1,155 @@
+# -*- coding: utf-8 -*-
+#######
+# actinia-core - an open source REST API for scalable, distributed, high
+# performance processing of geographical data that uses GRASS GIS for
+# computational tasks. For details, see https://actinia.mundialis.de/
+#
+# Copyright (c) 2016-2018 Sören Gebbert and mundialis GmbH & Co. KG
```
neteler, line 7:
```suggestion
# Copyright (c) 2023 Lina Krisztian and mundialis GmbH & Co. KG
```

tests/test_raster_import_pixellimit.py
```diff
@@ -0,0 +1,155 @@
...
+except Exception:
+    from test_resource_base import (
+        ActiniaResourceTestCaseBase,
+        URL_PREFIX,
+        additional_external_data,
+    )
+
+__license__ = "GPLv3"
+__author__ = "Lina Krisztian"
+__copyright__ = "Copyright 2016-2021, mundialis GmbH & Co. KG"
```
neteler, line 45:
```suggestion
__copyright__ = "Copyright 2023, mundialis GmbH & Co. KG"
```

tests/test_raster_import_pixellimit.py
```diff
@@ -0,0 +1,159 @@
...
+        self.create_new_mapset(self.tmp_mapset, location_name=self.location)
+
+    def tearDown(self):
+        # delete mapset
+        self.delete_mapset(self.tmp_mapset, location_name=self.location)
+        super(ImportRasterLayerPixellimitTestCase, self).tearDown()
+
+    def test_pixellimit_not_reached(self):
+        """
+        Test import of raster, for which pixellimit is not reached
```
mmacata, line 69: Suggestion to remove the negation (my first thought was that the test is meant to fail because of that even):
```suggestion
    def test_pixellimit_allowed(self):
        """
        Test import of raster, for which pixellimit is not reached and therefore allowed
```

tests/test_raster_import_pixellimit.py
```diff
@@ -0,0 +1,159 @@
...
+        self.waitAsyncStatusAssertHTTP(
+            rv,
+            headers=self.admin_auth_header,
+            http_status=200,
+            status="finished",
+        )
+
+    def test_pixellimit_reached(self):
+        """
+        Test import of raster, for which pixellimit is reached
```
mmacata, line 108: Suggestion to add the negation here (again my first thought was the opposite of what happens):

```suggestion
    def test_pixellimit_not_allowed(self):
        """
        Test import of raster, for which pixellimit is reached and therefore not allowed
```

src/actinia_core/processing/actinia_processing/ephemeral_processing.py
```diff
@@ -1203,6 +1204,154 @@ def _cleanup(self):
...
+                    )
+
+    def _check_pixellimit_rimport(self, process_executable_params):
+        """Check the current r.import command against the user cell limit.
+
+        Raises:
+            This method will raise an AsyncProcessError exception
+
+        """
+        rimport_inp = [x for x in process_executable_params if "input=" in x][
```
mmacata, line 1223: At this point, an importer command would already be translated to `r.import`, right? So the following would be parsed correctly because it is already transformed?:
```
 {
            "id": "importer_1",
            "module": "importer",
            "inputs": [
                {
                    "import_descr": {
                        "source": "https://raw.githubusercontent.com/mmacata/pagestest/gh-pages/bonn.geojson",
                        "type": "vector"
                    },
                    "param": "map",
                    "value": "polygon"
                }
            ]
        },
```

(Maybe add an importer PC to the tests below?) :upside_down_face:

src/actinia_core/processing/actinia_processing/ephemeral_processing.py
```diff
@@ -1203,6 +1204,154 @@ def _cleanup(self):
...
+        rastersize_x = int(rastersize_list[0])
+        rastersize_y = int(rastersize_list[1])
+        rastersize = rastersize_x * rastersize_y
+
+        # if different import/reprojection resolution set:
+        rimport_res = [
+            x for x in process_executable_params if "resolution=" in x
+        ]
+        res_val = None
+        # If raster exceeds cell limit alreaday in original resolution, next part can be skipped
```
mmacata, line 1285:
```suggestion
        # If raster exceeds cell limit already in original resolution, next part can be skipped
```

mmacata (approved):
Thanks, well done!

Please rename the tests like suggested or in another way, a negation in the title makes me expect a 400.
Other than that I "only" have ideas for more tests like using the importer, testing the same with ephemeral processing, ... but no blocker for me to be merged like this because tests for two cases exist.

src/actinia_core/processing/actinia_processing/ephemeral_processing.py
linakrisztian: Yes it is already transformed at this point. I also added a test for the importer

PR #498, opened 2023-12-12
Add tests via worker

mmacata:
This PR adds the command `make workertest`. It will run tests by writing processes to redis queue and starts a worker which will process it. See also https://github.com/actinia-org/actinia-core/issues/422 .

Possible TODOs later:
- Add test markers to only run tests relvant for worker execution (only asynchronous jobs are conducted by worker)
- Enhance tests further to test more queue types
- ...


docker/actinia-core-tests/actinia-test-worker-usedby-worker.cfg
```diff
@@ -0,0 +1,31 @@
...
+number_of_workers = 1
+redis_queue_server_url = localhost
+worker_queue_prefix = job_queue
+
+[LOGGING]
+log_level = 1
+
+[MISC]
+tmp_workdir = /tmp/actinia_core/workspace
+download_cache = /tmp/actinia_core/workspace/download_cache
```
anikaweinmann, line 31: Sind die beiden Configs nicht gleich?

anikaweinmann (approved):

docker/actinia-core-tests/actinia-test-worker-usedby-worker.cfg
mmacata: Yes they are equal - as preparation to have tests where they could differ.

PR #506, opened 2024-04-05
Remove python dependency

mmacata:
After update to alpine 3.19, the python package `jose` is no longer installed. After installing it, an error appears inside. As it is only used here to catch an exception which is then handled as any other exception, I suggest to remove this dependency.


anikaweinmann (approved):
Thank you

PR #535, opened 2024-06-20
Lint

mmacata:
This PR
- updates code to be valid for current black version
- adds "on pull_request" to lint check (before it was only on push and would not be visible in PR overview with only one commit)
- fixes other minor lint errors


anikaweinmann (approved):
Thank you :)

Issue #557, opened 2024-09-27
Support paths (and similar strings) in stdin


ninsbl:
It would be nice to be able to extract any string from module stdout and inject it to input to other parameters.
In particular, I am tinking of paths e.g. from `g.tempfile`, like e.g. here:

```JSON
{
                    "list": [
                        {
                            "flags": "d",
                            "id": "g_tempfile",
                            "inputs": [
                                {
                                    "param": "pid",
                                    "value": "123456789"
                                }
                            ],
                            "module": "g.tempfile"
                        },
                        {
                            "flags": "g",
                            "id": "g_region",
                            "inputs": [
                                {
                                    "param": "raster",
                                    "value": "elevation"
                                }
                            ],
                            "module": "g.region"
                        },
                        {
                            "flags": "g",
                            "id": "r_univar",
                            "inputs": [
                                {
                                    "param": "map",
                                    "value": "elevation"
                                },
                                {
                                    "param": "output",
                                    "value": "g_tempfile::stdout/temporary_file.txt"
                                }
                            ],
                            "module": "r.univar"
                        }
                    ],
                    "version": "3"
                }
```

`g_tempfile::stdout` above process chain is not parsed.
Would it be sufficient to add e.g. `/` here: https://github.com/actinia-org/actinia-core/blob/0e3db9e6131c8481ceef66237ccef0d546bcb39c/src/actinia_core/core/common/process_chain.py#L66

I am happy to create a PR.

PR #558, opened 2024-09-27
stdin: add path-sep to stdin-splitting

ninsbl:
Addresses #557 

Looking at adding a useful test...


mmacata:
Do you want this to be merged or add tests before?

ninsbl:
I think the right course of action is to add a test first... Will do that ASAP

PR #585, opened 2024-12-11
Use valkey instead of redis - functional adjustments

mmacata:
See https://github.com/actinia-org/actinia-core/issues/573


linakrisztian:
g83 tests fail, cause `valkey` is only supported for alpine `v3.20` onwards. These tests will be fixed in a separate PR.

PR #586, opened 2024-12-11
ruff D301

mmacata:
D301 [*] Use `r"""` if any backslashes in a docstring

Initially this was a test for the post-pr-review workflow. Now keeping it but including fixes.


mmacata:
Closing this. Would be nice to have to get rid of some ruff ignore rules though.

PR #600, opened 2025-04-04
Update actinia-core image for g83 tests

linakrisztian:
Fix failing g83 tests from #585 :
- valkey is only supported for alpine v3.20 onwards -> build new actinia core image with g83 and alpine 3.21, see https://github.com/actinia-org/actinia-core/issues/573#issuecomment-2775756651
- fix tests for `geodata_download_importer`: checking mimetypes with `magic` sometimes yields not  expected mimetype. E.g. a zip should yield `application/zip`, but sometimes returns `application/octet-stream`. This probably occurs due to some bug in `libmagic`, see also [here](https://github.com/ahupp/python-magic/blob/0e977c98649b4ed4533a7700fbc5fea6895bfe71/magic.py#L102). To avoid this, first check mimetype with [`mimetypes` module](https://docs.python.org/3/library/mimetypes.html). If format is not supported, check with `magic`.
- add `application/json` as valid mimetype

Additional changes: linting (JSON: no mixed spaces and tabs, markdown-linting, ruff: fixed [if-key-in-dict-del (RUF051)](https://docs.astral.sh/ruff/rules/if-key-in-dict-del/#if-key-in-dict-del-ruf051), add exception for [os-listdir (PTH208)](https://docs.astral.sh/ruff/rules/os-listdir/#os-listdir-pth208)) + update pre-commit-config with latest version of linting-github-workflow


mmacata (changes requested):
Thanks for the fix!

For the CHANGELOG.md - I would prefer to add an exception at the top of the file to have `*`allowed instead of `-` - When the release notes are generated, they use `*` . If we change to `-` we always need to adjust this. What do you think?

linakrisztian:
> Thanks for the fix!
> 
> For the CHANGELOG.md - I would prefer to add an exception at the top of the file to have `*`allowed instead of `-` - When the release notes are generated, they use `*` . If we change to `-` we always need to adjust this. What do you think?

Good idea. I just noticed, these changes came from `mdformat` linter, but here we use `markdownlint`, which allows `*`. So no exception needed

mmacata (approved):

PR #601, opened 2025-04-04
Remaining changes for switch of redis to valkey

linakrisztian:
NOTE: first merge #600 (linting + fix tests) + and #602 (semantic adjustment)

- update remaining usage of redis modules with valkey
- remove redis-config and add valkey-config


src/actinia_core/core/redis_lock.py
```diff
@@ -48,9 +48,9 @@ class RedisLockingInterface(object):
     # Return 1 for success and 0 for unable to acquire lock because
     # resource-lock already exists
     lua_lock_resource = """
-    local value_exists = redis.call('EXISTS', KEYS[1])
+    local value_exists = server.call('EXISTS', KEYS[1])
```
mmacata, line 51: Is this a typo? Or is `server` defined somewhere else? Also some occurances below.

mmacata (approved):
Looks all good except of the replacement with `server` - is this intentional?

src/actinia_core/core/redis_lock.py
linakrisztian: No, as I understood from the documentation, that is how it's called with valkey, see also section "Interacting with Valkey from a script" [here](https://valkey.io/topics/eval-intro/)

mmacata: but there must be something like `from valkey import server` or similar, no?

linakrisztian: No I don't think so. It's the LUA code which passed with `local_value_exists`. Example of this is also given [here](https://valkey-py.readthedocs.io/en/latest/lua_scripting.html).
And as I understood right from here https://github.com/valkey-io/valkey/issues/136 , we could either use `redis` or better `server`, thus `redis` ist just supported for backwards compatibility

PR #632, opened 2025-06-17
docker/actinia-core-alpine/start.sh: fix missing --text in grass call

neteler:
- fix missing `--text` in `grass` call (reported by @LorenzoBottaccioli)
- comment line break cosmetics


mmacata (approved):
Thanks!

PR #682, opened 2026-01-06
tests_with_kvdb.sh: fix comparison in shell

neteler:
Comparison in shell is done with single `=` to be portable.


mmacata (approved):
merci !

PR #685, opened 2026-01-08
Catch error if resource does not exist

mmacata:
Currently, actinia-core returns a 
>500 Internal Server Error

when a resource does not exist ( e.g. for http://127.0.0.1:8088/api/v3/resources/actinia-gdi/resource_id-obviously_wrong_id.

This PR fixes this and returns a `400`
```json
{
    "message": "Resource does not exist",
    "status": "error"
}
```

### DIFF
--- a/src/actinia_core/rest/resource_management.py
+++ b/src/actinia_core/rest/resource_management.py
@@ -206,12 +206,14 @@ class ResourceManager(ResourceManagerBase):
             )
 
         if resp_data is not None:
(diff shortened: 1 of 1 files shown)


mmacata:
I just realized that this would be a breaking change, so this needs more discussion.

mmacata:
Updated this PR to keep current behaviour (status code 400).
Opened [issue](https://github.com/actinia-org/actinia-core/issues/692) to discuss this further to not block the bug fix.

PR #732, opened 2026-06-02
remove setuptools from requirements

ninsbl:
setuptools is (at least should not be) not a runtime dependency, however with a specific version pinned, it causes a version conflict with actinia-grassdata-management-plugin==1.0.0
Therefore I suggest to remove it from dependencies (also from other plugins).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "actinia-core"
-version = "7.1.1"
+version = "7.1.2"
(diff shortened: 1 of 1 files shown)


ninsbl:
When you agree, and after merge, a new patch release may be necessary to make sure that pip install works again...

mmacata:
Yes fine with this change. Just reset the version, this is automatically done after a release (https://github.com/actinia-org/actinia-core/actions/workflows/update-version.yml)

mmacata (approved):

PR #758, opened 2026-09-15
Enable Dependency-Track reporting in SBOM workflow

TaniaGithub0401:
This PR enables Dependency-Track reporting for the existing SBOM vulnerability workflow in actinia-core.

The generated Docker SBOM is sent to the `actinia-core` project in Dependency-Track, where vulnerabilities can be analyzed and the corresponding CycloneDX VEX information can be managed.

The workflow uses the reusable Dependency-Track integration from `mundialis/github-workflows`.

The repository secret `DEPENDENCY_TRACK_API_KEY` needs to be configured before the workflow runs on `main`.


mmacata:
Is this contained in https://github.com/actinia-org/actinia-core/pull/760 ?

TaniaGithub0401:
Yes, exactly. I was implementing it step by step, so I opened PRs for the parts that were already working while continuing with the remaining changes until reaching the final implementation.

